PR #2, opened 2025-12-08
Add initial production backend: /api/upload + intake base row logic

Ye5888:
- Updated write_base_row in sheets_sync.py to write user input into sheets
- Added API endpoint in main.py for user input along with input validation


Kevin-TCUS (changes requested):
Per our Slack convo: logic collision in sheets_sync.py. Let's shift the parser range to M:AC to protect the Motivation column.

Kevin-TCUS (approved):
Looks perfect. Thanks for the quick fix on the column mapping!

Issue #37, opened 2025-12-19
[Parser] Improve Location Extraction Heuristics


Kevin-TCUS:
**Context**
`extract_location()` checks the first 5 lines and treats any line containing a comma as a location:
```python
for line in candidate_lines:
    parts = line.split(",")
    if len(parts) >= 2:
        return [line.strip()], 1.0
```

**Problem**
This is too weak and causes false positives on titles like: "Software Engineer, Backend Developer" or "Manager, IT".

**Task**
Make location extraction more specific than "comma exists".

**Suggested approach (keep it lightweight)**
1.  **Expand search scope:** Check the first 10–15 lines, but **exclude** lines containing email/phone/URLs.
2.  **Require a stronger match:**
    * **City, ST:** Match `Word, ST` or `Word ST` where `ST` is in the **official set of 50 US state codes** (+DC). Do not accept just any two capitalized letters (avoids "Manager, IT").
    * **Keywords:** Explicitly look for `Remote` or `Hybrid`.
3.  **Filter obvious noise:**
    * Ignore lines containing common title/role keywords (`engineer`, `developer`, `manager`, `intern`, `Inc.`, `LLC`) when they don’t match a strict state code.

**Acceptance criteria**
* [ ] Parser does not label job titles (e.g., "Developer, Python", "Manager, IT") as a location.
* [ ] Parser successfully extracts standard `City, ST` formats using valid US state codes.
* [ ] Parser successfully extracts `Remote` / `Hybrid` when present near the top.

PR #38, opened 2025-12-21
Improved location extraction heuristics and fixed false positives

Ye5888:
- Ensured locations matched a state code (NY, DC, etc)
- Filtered out lines containing URLs, emails, phone numbers, or job titles
- Expanded the scope from 5 lines to 15 lines

### DIFF
--- a/backend/parser.py
+++ b/backend/parser.py
@@ -1,6 +1,7 @@
 from nameparser import HumanName
 from names_dataset import NameDataset
 import re
+import us
 from typing import List, Dict, Tuple, Any
 
 nd = NameDataset()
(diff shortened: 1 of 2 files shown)


Kevin-TCUS (changes requested):
Hi @Ye5888 — thanks for taking this on. Overall direction here is good: using the us library for state validation and filtering out URL/email/phone noise is the right approach, and this aligns well with Issue #37.

I’m going to request changes before merging so this is safe and production-ready. This is very normal in PR-based workflows — the goal of review is to catch edge cases and completeness issues before they land on main.

Here are the items to address:
	1.	Code hygiene / duplicate assignment
In extract_location(), candidate_lines is defined twice (first with a 5-line limit, then immediately overwritten with a 15-line limit). Please remove the old 5-line assignment so there’s no dead code and the final logic is clear.
	2.	ZIP / extra token handling (common real-world input)
The current check state_candidate in state_codes is too strict. For a very common header like Raleigh, NC 27601, parts[1] becomes "NC 27601", which causes us to reject a valid location.
Suggested fix: strip extra tokens after the state code, e.g.
state_candidate = parts[1].strip().upper().split()[0].
This makes the parser more defensive against messy but realistic resume data.
	3.	Logic ordering: don’t drop valid locations
Right now the job-keyword filter runs before the state-code check. Many resumes combine role + location on one line (e.g. Software Engineer | Raleigh, NC), so filtering on job titles first can cause valid locations to be skipped.
Suggested fix: attempt the City/State match first; only apply the job-keyword skip if there is no valid state-code match.
	4.	Dependencies (important for “real world” PRs)
This PR adds import us. In professional workflows, if a PR introduces a new dependency, the dependency file is updated in the same PR so main remains installable immediately after merge.
Please confirm whether us is already listed in backend/requirements.txt; if not, add it there as part of this PR.

Quick process note (just to normalize this): this is exactly how PRs usually work in real software teams — first pass improves things, review catches edge cases and completeness issues, author pushes a follow-up commit to the same PR, then we merge.

Once these are addressed, this should be ready to go. Thanks again — we’re very close.

Kevin-TCUS (approved):
Everything else looks solid:

us dependency added 

Logic ordering fixed 

ZIP/extra token handling fixed 

Approving now!

PR #45, opened 2026-01-09
Fix error codes and add 5MB PDF size constraint

### LINKED ISSUE #41: [Performance] Make upload endpoint non-blocking & Add Strict Validation
### Context / Related
* **Related to:** #19 (Frontend implementation)
* **Scope:** This ticket is **Backend Only** (Python/FastAPI). The frontend already handles client-side validation; this ticket adds server-side enforcement as a safety net.

### Summary
The upload endpoint should return immediately after storing the submission and queuing work. Right now, the request handler still opens the PDF with `fitz` and extracts text before responding, which slows the request and reduces throughput.

We also need to enforce strict file validation (size and type) *before* uploading to Drive to prevent junk data from entering the system.

### Current Behavior
* `POST /api/upload` performs PDF parsing inside the request handler.
* This blocks the API response.
* **Gap:** No strict validation exists for file type or size in the backend.

### Desired Behavior
1.  **Strict Validation (First):**
    * Check file size (Max 5MB).
    * Check file type (Must be PDF via extension AND magic bytes `%PDF`).
    * Return errors immediately (`400` or `413`) if invalid.
2.  **Fast Processing:**
    * Upload to Drive.
    * Write base row to Sheets.
    * Enqueue background task.
    * Return `200 OK`.
3.  **Background Work:**
    * All `fitz` parsing happens here.

### Acceptance Criteria
- [ ] **Validation (Size):** Read file bytes and check `len(contents)`. If > 5MB, return `413 Payload Too Large`.
- [ ] **Validation (Type):** Check that the filename ends in `.pdf` **AND** the first 4 bytes of the content are `%PDF`. If not, return `400 Bad Request`.
- [ ] **Remove Blocking Code:** Remove all `fitz` usage from the request handler.
- [ ] **Background Task:** Update the background task to handle:
    -   Downloading the file (if needed).
    -   `fitz` parsing and text extraction.
    -   Updating the Google Sheet.
- [ ] **Error Handling:** Background task catches exceptions (e.g., corrupt file structure) and logs them without crashing the server.

### Implementation Notes
* **⚠️ Critical - Stream Handling:** When you `await file.read()` to check validation, you consume the file stream. You must pass those *read bytes* to the Drive uploader (do not try to read the stream again, or you will upload an empty file).
* **Constants:** Use a named constant for the size limit so it is easy to change later:
  `MAX_UPLOAD_BYTES = 5 * 1024 * 1024  # 5MB`
* **Status Codes:**
    * `413` for size violations.
    * `400` for type violations.

### Test Plan (Local Dev)
- [ ] **Speed Check:** Upload a valid PDF and confirm immediate response.
- [ ] **Size Check:** Try uploading a file > 5MB. Confirm `413` error.
- [ ] **Fake PDF Check:** Rename a `.jpg` or `.txt` file to `.pdf` and upload it. Confirm `400` error (Magic Bytes check).
- [ ] **Parse Check:** Confirm valid PDFs are still parsed correctly in the background.

jriya186:
Fixes #41

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -152,39 +152,22 @@ async def upload_volunteer_application(
 
         if len(file_bytes) > MAX_FILE_SIZE_BYTES:
             raise HTTPException(
-                status_code=422,
+                status_code=413,
                 detail={
                     "status": "error",
-                    "code": "VALIDATION_ERROR",
-                    "fields": {"file": "Max file size is 5MB.",
-                               },
+                    "code": "PAYLOAD_TOO_LARGE",
+                    "message": "Max file size is 5MB.",
                 },
             )
 
-        # 5) Basic PDF sanity check
-        try:
-            doc = fitz.open(stream=file_bytes, filetype="pdf")
-        except Exception:
-            traceback.print_exc()
-            raise HTTPException(
-                status_code=422,
-                detail={
-                    "status": "error",
-                    "code": "VALIDATION_ERROR",
-                    "fields": {"file": "PDF parsing failed: no text extracted",
-                               },
-                },
-            )
-
-        pre_text = "\n".join([p.get_text("text") for p in doc])
-        doc.close()
-
-        if not pre_text.strip():
+        # 5) PDF Signature Validation
+        if len(file_bytes) < 4 or file_bytes[:4] != b"%PDF":
             raise HTTPException(
                 status_code=400,
                 detail={
                     "status": "error",
-                    "message": "PDF parsing failed: no text extracted",
+                    "code": "BAD_REQUEST",
+                    "message": "Uploaded file is not a valid PDF.",
                 },
             )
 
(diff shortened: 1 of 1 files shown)


Kevin-TCUS (changes requested):
Hey @jriya186 — thank you for putting this together, I really appreciate you taking this on.

As we discussed, I cleaned up/replaced the earlier tracking issue, and the current “source of truth” requirements for this work are now in Issue #41. Your PR is close, but a few changes are still needed so it matches #41 and our current API contract.

Changes needed before I can approve/merge:

1.	File size status code (important):
If the uploaded file is over 5MB, please return 413 Payload Too Large (instead of 422). This is the standard HTTP code for upload size limits, and it’s what the frontend is expecting as a safety-net.

2.	PDF type validation (strict, non-blocking):
Validate “real PDF” using both:

	•	filename ends with .pdf (case-insensitive), and
	•	the first 4 bytes of the file are the PDF signature: %PDF
  If either fails, return 400 Bad Request.

3.	Non-blocking requirement:

Please remove all fitz usage from the request handler (POST /api/upload). Anything that opens/parses the PDF (including fitz.open(...)) should happen only in the background task, so the endpoint can return quickly after Drive upload + Sheet write + task enqueue.

Issue #42, opened 2026-01-09
[Optimization] Privacy-first parsing: Remove name/email inference dependencies


Kevin-TCUS:
### Summary
The current parser imports heavy dependencies (`names-dataset`) to "guess" the candidate's name and email from the PDF.
This is unnecessary because we already collect trusted Name/Email via the UI form. Furthermore, removing PII extraction logic makes it safer to run the parser on bulk datasets later, as we can focus purely on skills and experience without extracting sensitive identity data.

### Current Behavior
* Parser attempts to infer identity (Name, Email) from PDF content.
* Uses heavy dependencies (`names-dataset`) that consume resources.

### Desired Behavior
* **Stop extracting PII:** The parser should **not** attempt to find names or emails.
* **Blank Columns:** The `parsed_name` and `parsed_email` fields in the output should be `null` or empty strings.
* **Focus on Skills:** Keep the existing skill/experience extraction logic exactly as is.

### Acceptance Criteria
- [ ] **Remove Heavy Deps:** Remove `names-dataset` and `nameparser` from `requirements.txt` and the code.
- [ ] **Disable PII Extraction:** Delete logic that extracts name/email.
- [ ] **Clean Output:** Ensure the parser returns `None` (or empty) for name/email fields. (Do *not* backfill them with UI data; keep the "Parser" columns honest).
- [ ] **Preserve Skills:** Ensure existing skill extraction logic remains untouched.

### Test Plan (Local Dev)
- [ ] **Normal Upload:** Upload a standard resume.
    - [ ] Confirm the Sheet row shows **valid data** in the "UI Name/Email" columns (from the form).
    - [ ] Confirm the Sheet row shows **empty/blank** in the "Parsed Name/Email" columns.
    - [ ] Confirm "Parsed Skills" are still populated correctly.

PR #46, opened 2026-01-14
Feature/remove name parsing

Ye5888:
Issue #42 : Remove names-dataset and nameparser from requirements.txt and the code; Delete logic that extracts name/email and returns empty for those fields; Skills parsing not changed

### DIFF
--- a/backend/parser.py
+++ b/backend/parser.py
@@ -1,10 +1,7 @@
-from nameparser import HumanName
-from names_dataset import NameDataset
 import re
 import us
 from typing import List, Dict, Tuple, Any
 
-nd = NameDataset()
 
 def _get_lines(text: str) -> List[str]:
     return [line.rstrip() for line in text.splitlines()]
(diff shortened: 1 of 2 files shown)


Kevin-TCUS (changes requested):
Hey @Ye5888 — thank you for this PR. 🙏 This is a really strong contribution.

The core change is exactly what we need for Issue #42: removing names-dataset / nameparser and disabling name/email inference so we’re not extracting PII from resumes.

Two small requests before I can approve:
	1.	Confidence scores for blank PII
Right now name/email are blank (correct), but the confidence values are still 1.0. Could you please set confidence to 0.0 when the value is blank? That keeps the “parsed” columns honest and avoids sending a misleading signal.
	2.	Location work as a follow-up PR
I also noticed you improved location extraction and added the us dependency. The location changes look thoughtful, and I definitely want to keep that work, but it expands beyond the scope of #42.
Could you revert the location-related changes (in the Python file) and remove us from requirements.txt in this PR, so we can merge the privacy change cleanly?

After this merges, please open a second PR that reintroduces your location improvements + the us dependency. I’ll review that follow-up quickly, and we’ll get your location work landed the right way.

Once the confidence fix + location reverts are in, I’m happy to approve. Thanks again, seriously great work.

Kevin-TCUS:
Hey @Ye5888 — thanks again for the work here. Quick clarification after checking main:
	•	The location improvements + the us dependency are already merged into main (and that’s fine, we want to keep them).
	•	This PR is tracked against Issue #42, so the remaining scope should be only: remove name/email inference + remove nameparser + names-dataset.

Because this branch also contains earlier location commits, GitHub is now showing merge conflicts. Could you please update the PR so it cleanly applies on top of current main?

What I’m asking you to do:
	1.	Resolve the merge conflicts by keeping main’s location logic and keeping us (since that is already the canonical version).
	2.	In this PR, make sure we remove nameparser and names-dataset from requirements.txt and delete/disable any name/email extraction.
	3.	For the parser output: if name / emails are blank, please set confidence = 0.0 (so blank values don’t show 1.0 confidence).

Once conflicts are resolved and the PR matches #42, I’m happy to approve. 🙏

Kevin-TCUS (approved):
LGTM! thanks for the quick turnaround on this, Song.

I verified the changes:
* `confidence: 0.0` is set correctly for the blank name/email fields (keeping the data honest).
* `names-dataset` is gone (huge win for the Pi memory usage).
* `us` dependency is preserved, so the location logic is safe.

This is exactly what we needed for #42. Merging now. Good luck with classes this week!

PR #47, opened 2026-01-14
Feat/intake form frontend fix

jliu233:
Revised intake form frontend code to align with the figma design file


Kevin-TCUS (changes requested):
Hi Jiaqi, as discussed. Please use for the 'intake form' page the Google Studio code as an example, but ensure we are building per the API spec - as I do no know if Google generated code fully does. It looks really nice, but we must check it works with the backend as desired and complies with the API spec.

PR #58, opened 2026-01-26
Sheets schema and locations guard

### LINKED ISSUE #50: Column Mismatch Between Parser Output and Google Sheets Schema
Recent updates to the resume parser removed name and email extraction from resume PDF documents, as these fields are now collected directly from the UI. However, the Google Sheets schema and sync logic have not yet been updated to reflect this change.
As a result, the parser output no longer aligns with the expected sheet columns, leading to a column mismatch during updates and potential data misalignment.

**Current behavior:**
The parser no longer produces:

- parsed name
- parsed email
- related confidence fields

The Google Sheets update logic still expects those fields to exist.
This leads to:

- column shifting
- data ending up in the wrong columns
- parser confidence being calculated using fields that are no longer parsed

**Expected behavior**

- Name and email should be treated as UI-owned fields and not written by the parser. (already done)
- The parser output and Google Sheets columns should be updated so they stay in sync.
- Overall parser confidence should be computed only from fields that are still actively parsed (e.g. location, skills, education, experience).

**Proposed next steps**

- [ ] Update update_resume_in_sheet() to stop writing parsed name and email values and realign the remaining columns.
- [ ] Adjust parser confidence calculation to reflect the current parser outputs.

jriya186:
Updated comments in sheets_sync to reflect correct column scheme in google sheets. Also added a guard in location extraction that prevents a crash when location is empty or malformed. Fixes #50

### DIFF
--- a/backend/parser.py
+++ b/backend/parser.py
@@ -91,9 +91,12 @@ def extract_location(text: str) -> Tuple[List[str], float]:
     
         parts = line.split(",")
         if (len(parts) >= 2):
-            state_candidate = parts[1].strip().upper().split()[0]
-            if (state_candidate in state_codes):
-                return [line.strip()], 1.0
+            tokens = parts[1].strip().upper().split() # safely tokenize state segment - might be empty or malformed
+            if tokens:
+                state_candidate = tokens[0]
+                if state_candidate in state_codes: # validate against known US state abbreviations
+                    return [line.strip()], 1.0
+    
         if any(re.search(rf'\b({job})\b', line, re.IGNORECASE) for job in job_keywords):
                 continue
     return [], 0.0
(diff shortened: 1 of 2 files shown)


Kevin-TCUS (approved):
PR Review: Approved

Great catch on the column overlap.

Schema Fix: Confirmed that Motivation occupies Column L (index 11), so shifting the Parser Output to start at Column M (index 12) is the correct fix to prevent data collision.

Location Guard: The explicit check for if tokens: in parser.py is a good defensive addition. Without it, malformed location strings (like those in our synthetic test set) would crash the backend with an IndexError.

This looks good to go. Merging this now will clear the path for the upcoming intake form changes

PR #62, opened 2026-02-09
Implemented skill resolution logic in resolver.py

sharzoom:
Implement the resolution logic for extracted skills, including normalization, alias mapping, and coverage statistics calculation.

### DIFF
--- a/backend/resolver/resolver.py
+++ b/backend/resolver/resolver.py
@@ -1,5 +1,6 @@
-from typing import Dict, Optional
+from typing import Dict, Optional, List, Set, Tuple
 
+from .normalize import normalize_key
 from .schemas import (
     ExtractedProfileV1,
     ResolvedProfileV1,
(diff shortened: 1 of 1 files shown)


Kevin-TCUS (commented):
This is a really strong implementation. It perfectly adheres to the “pure function” requirement and the logic matches the spec closely.

What looks great:

Pure logic only, no I/O or external calls.

Order-preserving deduplication for resolved skills using a seen set while iterating, which keeps first-seen order stable.

Unknowns are deduped by normalized key while preserving the first raw string encountered, which is exactly what we want for auditability.

Coverage math is computed over unique normalized concepts, which matches the spec perfectly.

One dependency note:
This PR’s correctness depends on normalize_key producing distinct keys for important tokens (C++ vs C# vs C, .NET, etc.). So I will hold off on hitting the merge button until the normalization fix in PR #61 lands.

After #61 is updated, I’m happy to approve this one. Thanks again for knocking this out!

Kevin-TCUS (approved):
This is a fantastic implementation, Fathima! The logic matches the spec perfectly, the deduplication is spot on, and the architecture is exactly what we needed. I am thrilled to approve and merge this. Thank you so much for your hard work on the core resolver!

Issue #63, opened 2026-02-13
[Parser] Improve experience extraction


jriya186:
**Description:**
The parser is currently extracting Work Experience incorrectly because it matches any occurrence of the word “experience” in the resume text. This causes content from sections like Professional Summary (e.g., “5+ years of experience in Python”) to be incorrectly identified as Work Experience.
Project Experience is also not consistently parsed due to the same issue. The current implementation does not strictly track section boundaries and relies on broad keyword matching.

**Root Cause:**
Section detection uses loose patterns such as **r'experience',** which match partial lines and inline mentions rather than explicit section headers. As a result, the parser cannot reliably distinguish between a section header and a sentence containing the word “experience.”

**Expected Behavior:**
The parser should detect only explicit section headers and track section boundaries until the next valid header. It should not match inline occurrences of keywords within summary or bullet text.

**Solution:**
Implement strict section tracking with anchored regex patterns for headers (e.g., matching full header lines only). Add explicit section start and end detection and ensure Work Experience and Project Experience are handled as separate, clearly defined sections.

**Impact:**
This change will reduce false positives, improve parsing accuracy for Work and Project Experience, and increase overall parser reliability.

jriya186:
I’ve updated the section detection logic so that Work Experience is no longer triggered by any occurrence of the word “experience” within the resume text. The parser now only begins capturing content when it detects an explicit section header (e.g., “Experience,” “Work Experience,” “Professional Experience”) using anchored matching. This prevents summary lines such as “5+ years of experience…” from being incorrectly parsed as work history.
During testing, I observed that multi-column resumes and certain template-based layouts (e.g., Canva or sidebar designs) result in scrambled text extraction from the PDF parser. In these cases, the actual job entries may appear before the “Experience” header in the extracted text stream. As a result, strict section-based state tracking fails because the header does not precede the content logically.
To handle these layouts, we will need a fallback heuristic that detects experience entries based on strong patterns (e.g., job title + organization + date/location patterns) when section tracking returns empty. This hybrid approach should preserve precision for well-structured resumes while improving recall for multi-column or template-based formats. Let me know if that works or if there are better suggestions for this.

PR #64, opened 2026-02-20
Experience parsing fix

jriya186:
Improves #63. Improved experience extraction to not parse text as experience unless its under a header. Stops from false parsing from summary or skills sections with the word 'experience'.


Kevin-TCUS (changes requested):
Awesome work, Riya! This is a great architectural shift. It looks like now we strictly create an experience section if the line is exactly: Experience, Work Experience, Professional Experience, or Employment (and optionally with a colon). This is perfect for killing those false positives!

Also, great eye catching that math/slicing bug in the project confidence calculation (len(lines) instead of slicing twice). That was a sneaky one!

I just have one quick required fix for the PR, and one optional idea:

Required: The Section Boundary Checker

You smartly added :? to your local patterns to catch colons, but the global regex string in _is_section_header still strictly ends with $. Please update _is_section_header to end with :?$ so the global boundary check doesn't fail to recognize those headers with colons!

Optional: Catching a few more (feel free to push to a new ticket)

Since we are anchored now, we could optionally catch a few more valid edge cases by adding these to work_patterns:

r'^career\s+history:?$'

r'^work\s+history:?$'

r'^relevant\s+experience:?$'

Let me know when you push the boundary checker fix and I'll get this merged!"

Kevin-TCUS:
Closing this out since we successfully merged the clean branch over in #66!

Thanks again for untangling the commits and getting this parser upgrade shipped, @jriya186 !

PR #65, opened 2026-02-22
feat: initialize benchmark harness and link to parser

Ye5888:
Created benchmark.py for benchmark harness and grade parser's ability to pick up skills/location/other info


Kevin-TCUS (changes requested):
This is a phenomenal piece of work. You nailed the CLI contract, the adapter pattern, and the fail-safe error handling. Having a repeatable benchmark harness is a massive milestone for Libelle.

I have three required scoring tweaks to make results truly deterministic and accurate for v1, plus a couple optional improvements if you have time.

Required fixes for merge (blockers)


Make skills examples deterministic
Python set iteration order can vary between runs. To keep the benchmark repeatable, please sort TP/FP/FN sets before slicing them into example lists.
Don’t penalize empty golden locations
If the Golden JSON has no country, but the parser predicts one, that should be a false positive only. Right now it also records a false negative, which artificially tanks recall.
Fix the naive location split for US-style locations
_split_location(“Raleigh, NC”) currently treats “NC” as the country, which breaks the strict country-matching rule. For v1, a minimal fix is to treat a trailing two-letter token as a US state code and map country to “united states”.

Optional polish (nice-to-haves / v2)


Tracebacks in run_log.json: you print tracebacks to console, but only store str(e) in JSON. Saving the formatted traceback into the error array would make debugging regressions much easier.
Sort examples.md by worst offenders: instead of the first 5 FP/FN rows, consider sorting by fp_count / fn_count descending so the biggest failures surface first.
Also noticed the run log shows Python 3.14, love the bleeding edge energy. Just make sure we aren’t relying on anything that will break on 3.10/3.11 in CI.

Ping me when the three blockers are addressed and I’ll merge.

Kevin-TCUS (changes requested):
Hey @Ye5888  — this update looks strong. You addressed the three requested blockers well: deterministic skill examples, empty golden location handling, and US-style state parsing. I also appreciate the extra polish on traceback capture and the Markdown sorting.

The benchmark logic looks ready. I just have two quick repo cleanup items before merge:
	1.	Keep a single canonical runner file
Please rename the updated script to benchmark.py and remove benchmark_new.py.
	2.	Remove generated run artifacts from the PR
Please remove the timestamped backend/benchmarks/runs/... folder and its generated files from this PR. The harness should generate those locally at runtime, but we don’t want to commit one specific run output into main. (though it was cool to see the results in the csv!)


Once the duplicate script and generated run outputs are removed, I’m happy to merge.

Kevin-TCUS (approved):
Looks good to merge. Thanks for cleaning up the duplicate runner and removing the generated run artifacts. This gives us a strong first benchmark baseline for parser evaluation.

PR #66, opened 2026-03-06
fixed exp extraction

### LINKED ISSUE #63: [Parser] Improve experience extraction

jriya186:
This PR reintroduces the experience extraction fix on top of the current main branch.
The previous branch included several commits that originated from earlier branches (golden-dataset, dataset additions, and parser updates) because those branches were stacked during development. Since those commits have already been merged into main, this new PR was created from a clean branch to isolate only the intended change.
The update improves experience parsing by ensuring the parser only begins extracting work experience when a valid section header (e.g., “Experience”, “Work Experience”, “Professional Experience”) is detected. This prevents false positives where the word “experience” appears in other sections such as summaries.
The section header detection has also been expanded to support additional common variants (e.g., “Career History”, “Work History”, “Relevant Experience”) and optional trailing colons. Fixes #63

### DIFF
--- a/backend/parser.py
+++ b/backend/parser.py
@@ -11,7 +11,12 @@ def _is_section_header(line: str) -> bool:
         return False
     s = line.strip()
     headers = [
-        r'^(summary|objective|contact|education|certifi|certificate|skills|work experience|work|experience|employment|projects|project experience|project|research|publications|awards|volunteer|honors|activities)$'
+
+        r'^(summary|objective|contact|education|certifi|certificate|skills|'
+        r'work experience|professional experience|experience|employment|'
+        r'career history|work history|relevant experience|'
+        r'projects|project experience|project|research|publications|'
+        r'awards|volunteer|honors|activities):?$'
     ]
     if re.match(headers[0], s.lower()):
         return True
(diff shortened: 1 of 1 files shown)


Kevin-TCUS (approved):
This is clean and aligned with our deterministic parsing goals.

Switching to anchored header patterns and using match() instead of search() eliminates the false positives we were seeing from summary lines. The expanded header vocabulary is a good balance between precision and recall.

I’ve reviewed the changes and don’t see any regressions relative to main.

Approving.

Nice tightening of the extraction logic.

PR #94, opened 2026-04-01
Updated aliases_v1

sharzoom:


Kevin-TCUS (approved):
Splitting these semantic variants out into the alias map was exactly the right architectural move! A few of these keys have spaces (which our normalizer strips, making them technically unreachable), but I'm going to merge this as-is so we can keep our momentum going. We can easily clean up those spaces in a follow-up sweep. Thank you again for such a meaningful contribution to Libelle!

PR #108, opened 2026-04-06
Fix: write_base_row() signature mismatch + submission_id column placement

jliu233:
### Summary

Fixes upload failure in the backend caused by a function signature mismatch between `main.py` and `sheets_sync.py`, and corrects `submission_id` being written to the wrong column in Google Sheets.

### Changes

* Removed unused `resume_id` parameter from `write_base_row(...)`
* Aligned `write_base_row(...)` signature with `main.py` call site
* Ensured `submission_id` is written to an append-only last column (not hardcoded `AD`)
* Added logic to dynamically detect / create `submission_id` column as the true last column
* Preserved existing sheet schema (no column shifting)

### Root Cause

`main.py` called:

```python
write_base_row(drive_file_id=..., ...)
```

But `sheets_sync.py` required:

```python
write_base_row(resume_id, drive_file_id, ...)
```

Resulting in:

```bash
TypeError: write_base_row() missing 1 required positional argument: 'resume_id'
```

### Result

* Upload flow completes successfully end-to-end
* PDF uploads to Drive
* Base row writes to Sheets
* Frontend receives success response
* `submission_id` correctly stored in last column (AI)

### Testing

* Submitted form via frontend
* Verified:

  * Drive upload success
  * Sheet row appended
  * `submission_id` appears in last column
  * No backend 500 error

### Closes

Closes #102

### DIFF
--- a/backend/sheets_sync.py
+++ b/backend/sheets_sync.py
@@ -56,50 +56,6 @@ def _drive_link(file_id: str) -> str:
 def _local_timestamp() -> str:
     return datetime.now(timezone.utc).strftime("%m-%d-%Y %H:%M:%S %Z")
 
-def _column_letter(column_number: int) -> str:
-    result = ""
-    while column_number > 0:
-        column_number, remainder = divmod(column_number - 1, 26)
-        result = chr(65 + remainder) + result
-    return result
-
-
-def _get_headers() -> List[str]:
-    header_values = sheet.values().get(
-        spreadsheetId=GOOGLE_SHEET_ID,
-        range=f"{SHEET_NAME}!1:1",
-    ).execute().get("values", [])
-    return header_values[0] if header_values else []
-
-
-def _ensure_submission_id_column() -> int:
-    """
-    Ensure `submission_id` exists as the last column header.
-    Returns the 1-based column number for submission_id.
-
-    Important:
-    - append-only
-    - never insert at column A
-    - never shift existing columns
-    """
-    headers = _get_headers()
-
-    if "submission_id" in headers:
-        return headers.index("submission_id") + 1
-
-    column_number = len(headers) + 1 if headers else 1
-    column_letter = _column_letter(column_number)
-
-    sheet.values().update(
-        spreadsheetId=GOOGLE_SHEET_ID,
-        range=f"{SHEET_NAME}!{column_letter}1",
-        valueInputOption="RAW",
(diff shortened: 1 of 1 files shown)


Kevin-TCUS (changes requested):
Hey @jliu233, great catch on fixing the update_resume_in_sheet signature for #106 alongside #102. The safer .get() handling also makes the parser writeback path more resilient.

However, I’m hesitant to approve this as-is because it bundles the critical bug fixes with a broader behavior change: the dynamic submission_id placement / header mutation logic.

For v0.1, I want to enforce scope tightly. The goal here is to sync the repo with the critical fixes needed to restore the working pipeline, not introduce additional runtime schema behavior. If something unexpected happens with the new header logic in production, reverting it would also roll back the critical bug fixes, which I want to avoid.

Could you please trim this PR so it includes only what is needed to resolve the two known bugs:
	•	write_base_row() signature alignment for #102
	•	update_resume_in_sheet() signature alignment for #106
	•	any minimal safe .get() changes directly required to make parser writeback stable

Please remove the submission_id dynamic column/header logic and anything else beyond those bug fixes from this PR.

I do not want more v0.1 feature work here. I just want the repo caught up to the proven bug fixes so we can merge, stabilize, and move on. Meaning: we will not change how / where submission Id is written in v0.1 as we will leave it behind with a brand new Google Sheet setup for V0.3

Kevin-TCUS (changes requested):
Hey @jliu233, thanks for updating this so quickly and removing the dynamic column logic!

It looks like there's a slight Git branch tangle happening. I see you opened PR #109 for the Issue #106 fixes, which is the perfect approach! However, the diff here in PR #108 still contains all of the changes for update_resume_in_sheet and the .get() refactoring.

Since PR #108 is meant to be a 1:1 sync with the production hotfix for Issue #102, we need this diff to show only the removal of resume_id from the write_base_row signature. Everything else in sheets_sync.py should look exactly as it currently does in main.

Git branch splitting can be tricky! The easiest way to resolve this on this branch (fix/remove-resume-id) is to:

Revert backend/sheets_sync.py back to exactly how it looks in the main branch.

Make the single edit to remove resume_id from write_base_row.

Commit and push to this branch (you may need to use a force push).

Once this PR is back to just that single line change, I'll approve and merge it immediately, and then I'll head over to review PR #109. Thanks for your patience as we get these cleanly separated!

PR #109, opened 2026-04-06
Fix: update_resume_in_sheet() signature mismatch in parser writeback

jliu233:
### Summary

Fixes background parser writeback failure caused by a function signature mismatch between `main.py` and `sheets_sync.py`.

### Changes

* Removed unused `resume_id` parameter from `update_resume_in_sheet(...)`
* Aligned function signature with current call site:

  ```python
  update_resume_in_sheet(parsed)
  ```
* Kept row lookup based on `drive_file_id` (no dependency on `resume_id`)

### Root Cause

`main.py` calls:

```python
update_resume_in_sheet(parsed)
```

But `sheets_sync.py` previously defined:

```python
def update_resume_in_sheet(resume_id: int, parsed: Dict[str, Any])
```

This caused the background job to fail with:

```bash
TypeError: update_resume_in_sheet() missing 1 required positional argument: 'parsed'
```

### Result

* Upload flow completes successfully
* Background parsing no longer crashes
* Parsed fields are correctly written back to the Google Sheet
* End-to-end pipeline (upload → base row → parse → writeback) is fully functional

### Notes

`resume_id` is not used anywhere in the function body, so removing it is safe and aligns with the current pipeline design that relies on `drive_file_id` for row matching. 

### Testing

* Submitted a volunteer application
* Verified:

  * Upload success response returned
  * No background job errors in logs
  * Parsed data populated in columns `M:AC`
  * Correct row updated using `drive_file_id`

### Closes

Closes #106

### DIFF
--- a/backend/sheets_sync.py
+++ b/backend/sheets_sync.py
@@ -56,50 +56,6 @@ def _drive_link(file_id: str) -> str:
 def _local_timestamp() -> str:
     return datetime.now(timezone.utc).strftime("%m-%d-%Y %H:%M:%S %Z")
 
-def _column_letter(column_number: int) -> str:
-    result = ""
-    while column_number > 0:
-        column_number, remainder = divmod(column_number - 1, 26)
-        result = chr(65 + remainder) + result
-    return result
-
-
-def _get_headers() -> List[str]:
-    header_values = sheet.values().get(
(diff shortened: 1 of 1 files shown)


Kevin-TCUS (changes requested):
See comments in Slack!

PR #116, opened 2026-04-08
fix extract_skills, achored header patters, expanded vocab

### LINKED ISSUE #81: extract_skills() — incorrect section detection and spurious content extraction
**Description**
extract_skills() in backend/main.py exhibits two distinct failure modes when tested against real resume PDFs:

False positive triggers: the section header detector fires on any line containing the word "skills", including sentences in summary or objective paragraphs (e.g. "Strong communication skills are essential"), causing non-skills content to be harvested as output.
False negatives: resumes using alternative header phrasing (Technical Proficiencies, Core Competencies, Technologies) are not detected at all, returning an empty skills list with confidence = 0.0.

**Root cause**
In parser.py:
# Fires on any sentence containing "skills", not just headers
if re.search(r'\bskills\b', line, re.IGNORECASE):
    in_skills_section = True
The function relies entirely on a single loose keyword match with no structural signals. The closed single-keyword vocabulary cannot handle natural variation in resume header phrasing.

**Impact**
Skills list is populated with random, inaccurate content with no indication of failure
confidence = 1.0 is returned even for completely incorrect extractions, making the score unreliable as a quality signal

**Suggested fix**
Replace the single-keyword heuristic with a predefined header vocabulary and anchored matching, consistent with the approach already used in extract_work_experience(). Longer term, consolidate section boundary detection into a single-pass segment_resume() so all extractors share one view of document structure.

jriya186:
Fixes #81 

### Summary
Rewrote extract_skills() to use anchored header patterns and _collect_section_lines() for consistent section boundary detection, replacing the loose re.search heuristic that caused false positives
Added job experience to work_patterns in extract_work_experience()
Updated _is_section_header() and _collect_section_lines() to strip trailing dashes and underscores before matching, fixing missed headers caused by decorative formatting (e.g. Skills ___________)
Expanded _is_section_header() vocabulary to include volunteer experience and job experience as valid stop signals

### What changed

**extract_skills()** — full rewrite using anchored patterns and _collect_section_lines()
**extract_work_experience()** — added job experience to start patterns
**_is_section_header()** — trailing decorator cleanup + expanded vocab
**_collect_section_lines()** — trailing decorator cleanup applied before start and stop matching

### Small note about benchmark performance:

The aggregate F1 dipped from 0.581 to 0.530 after these changes, but the per-resume breakdown makes clear this is an upstream extraction quality issue rather than a parser logic regression.

The resumes with clean text layers all score well — high_signal_01, high_signal_02, non_usa_01, non_usa_02 score consistently, with non_usa_02 hitting a perfect F1 of 1.0. The dip is driven entirely by resumes where fitz cannot produce reliable text:

- multi_col_01, multi_col_02 — multi-column layouts where fitz reads columns out of order, producing scrambled text that the parser can't segment correctly
- embed_link_02, low_signal_01 — return empty or near-empty text, so all extractors produce zero output regardless of parser logic
- low_signal_02 — text scrambling causes false positives where volunteer/experience lines are captured as skills

These failures are pre-existing and not introduced by this PR. For standard text-layer PDFs, the section boundary improvements in this PR produce meaningfully better results. The multi-column and image-based PDF issues are a separate upstream problem that will need to be addressed at the extraction layer.

### DIFF
--- a/backend/parser.py
+++ b/backend/parser.py
@@ -6,10 +6,13 @@ from typing import List, Dict, Tuple, Any
 def _get_lines(text: str) -> List[str]:
     return [line.rstrip() for line in text.splitlines()]
 
+def _clean_line(line: str) -> str:
+    return re.sub(r'[\s_\-=*]+$', '', line.strip())
+
 def _is_section_header(line: str) -> bool:
     if not line or len(line.strip()) == 0:
         return False
-    s = re.sub(r'[\s_\-=*]+$', '', line.strip())
+    s = _clean_line(line)
     headers = [
 
         r'^(summary|objective|contact|education|certifi|certificate|skills|'
@@ -31,12 +34,12 @@ def _collect_section_lines(lines: List[str], start_patterns: List[str], stop_whe
     capturing = False
     end_index = len(lines)
     for i, line in enumerate(lines):
-        cleaned = re.sub(r'[\s_\-=*]+$', '', line.strip())
+        cleaned = _clean_line(line)
         if not capturing and start_re.match(cleaned):
             capturing = True
             continue
         if capturing:
-            if stop_when_header and _is_section_header(cleaned):
+            if stop_when_header and _is_section_header(line):
                 end_index = i
                 break
             collected.append(line)
--- a/backend/parser.py
+++ b/backend/parser.py
@@ -9,14 +9,15 @@ def _get_lines(text: str) -> List[str]:
 def _is_section_header(line: str) -> bool:
     if not line or len(line.strip()) == 0:
         return False
-    s = line.strip()
+    s = re.sub(r'[\s_\-=*]+$', '', line.strip())
     headers = [
 
         r'^(summary|objective|contact|education|certifi|certificate|skills|'
         r'work experience|professional experience|experience|employment|'
+        r'job experience|'
         r'career history|work history|relevant experience|'
         r'projects|project experience|project|research|publications|'
-        r'awards|volunteer|honors|activities):?$'
+        r'awards|volunteer|volunteer experience|honors|activities):?$'
     ]
     if re.match(headers[0], s.lower()):
         return True
@@ -30,11 +31,12 @@ def _collect_section_lines(lines: List[str], start_patterns: List[str], stop_whe
     capturing = False
     end_index = len(lines)
     for i, line in enumerate(lines):
-        if not capturing and start_re.match(line.strip()):
+        cleaned = re.sub(r'[\s_\-=*]+$', '', line.strip())
+        if not capturing and start_re.match(cleaned):
             capturing = True
             continue
         if capturing:
-            if stop_when_header and _is_section_header(line.strip()):
+            if stop_when_header and _is_section_header(cleaned):
                 end_index = i
                 break
             collected.append(line)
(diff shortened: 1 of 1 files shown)


Kevin-TCUS (commented):
Hi Riya, this is incredibly clean work. Moving `extract_skills()` to the `_collect_section_lines()` architecture is exactly the kind of improvement the parser needed, and the anchored headers look great.

**My understanding of the old/new behavior is:**

**Old behavior:**
* `extract_skills()` used a looser heuristic to detect the skills section.
* It could trigger when "skills" appeared anywhere in a line, rather than only when the line behaved like a true section header.
* Once it started capturing, the stop logic was weaker, which made section bleed-through more likely when the next section header was not recognized cleanly.

**New behavior:**
* `extract_skills()` now uses anchored header patterns to identify the start of the skills section more intentionally.
* It now routes through `_collect_section_lines()`, which gives it the same section-boundary handling pattern used elsewhere in the parser.
* `_is_section_header()` / `_collect_section_lines()` now also handle decorative trailing characters more cleanly, which should help with headers like `Skills ________`.
* Expanding the stop-signal vocabulary also makes it less likely that adjacent sections bleed into extracted skills.

That feels like a much stronger architectural direction to me: stricter when deciding when to start, and more systematic when deciding when to stop.

**Before we merge, I have two small requests so the PR history captures the full context:**

**1. Could you add a short note to the PR description about the benchmark results?**
Your Slack analysis on the F1 dip (0.581 -> 0.530) was really helpful. Please add a brief summary that:
* The change improves section-boundary logic for standard text-layer PDFs.
* The aggregate dip appears tied mainly to multi-column / text-scrambled PDFs.
* This is an upstream extraction-quality issue, not just a parser-logic issue.

*That context will help future contributors understand why this is still a meaningful improvement.*

**2. One tiny code-cleanup question:**
I noticed the same trailing-decorator cleanup regex is now being applied in both `_collect_section_lines()` and `_is_section_header()`. Since those functions interact closely, could you either consolidate that cleanup into one place or add a short note on why you want it in both? I just want to make sure the behavior stays intentional and easy to reason about.

Once that’s updated, I should be in good shape to merge. Really nice work here.

jriya186:
> Hi Riya, this is incredibly clean work. Moving `extract_skills()` to the `_collect_section_lines()` architecture is exactly the kind of improvement the parser needed, and the anchored headers look great.
> 
> **My understanding of the old/new behavior is:**
> 
> **Old behavior:**
> 
> * `extract_skills()` used a looser heuristic to detect the skills section.
> * It could trigger when "skills" appeared anywhere in a line, rather than only when the line behaved like a true section header.
> * Once it started capturing, the stop logic was weaker, which made section bleed-through more likely when the next section header was not recognized cleanly.
> 
> **New behavior:**
> 
> * `extract_skills()` now uses anchored header patterns to identify the start of the skills section more intentionally.
> * It now routes through `_collect_section_lines()`, which gives it the same section-boundary handling pattern used elsewhere in the parser.
> * `_is_section_header()` / `_collect_section_lines()` now also handle decorative trailing characters more cleanly, which should help with headers like `Skills ________`.
> * Expanding the stop-signal vocabulary also makes it less likely that adjacent sections bleed into extracted skills.
> 
> That feels like a much stronger architectural direction to me: stricter when deciding when to start, and more systematic when deciding when to stop.
> 
> **Before we merge, I have two small requests so the PR history captures the full context:**
> 
> **1. Could you add a short note to the PR description about the benchmark results?** Your Slack analysis on the F1 dip (0.581 -> 0.530) was really helpful. Please add a brief summary that:
> 
> * The change improves section-boundary logic for standard text-layer PDFs.
> * The aggregate dip appears tied mainly to multi-column / text-scrambled PDFs.
> * This is an upstream extraction-quality issue, not just a parser-logic issue.
> 
> _That context will help future contributors understand why this is still a meaningful improvement._
> 
> **2. One tiny code-cleanup question:** I noticed the same trailing-decorator cleanup regex is now being applied in both `_collect_section_lines()` and `_is_section_header()`. Since those functions interact closely, could you either consolidate that cleanup into one place or add a short note on why you want it in both? I just want to make sure the behavior stays intentional and easy to reason about.
> 
> Once that’s updated, I should be in good shape to merge. Really nice work here.

Hi Kevin,
I have added a brief description of the benchmark performance in the PR description above and I consolidated the trailing decorator cleanup into a single helper `_clean_line()` so it's defined once and reused. `_is_section_header` calls it internally, and `_collect_section_lines` calls it for the start pattern match (since start_re.match has no cleanup of its own). For the stop check in `_collect_section_lines`, I'm passing the raw line directly to `_is_section_header` since it already handles cleanup internally so there's no duplication there. The behavior is identical, just cleaner to reason about now.

Kevin-TCUS (approved):
Thanks for making these final updates, @jriya186 . The added context regarding the benchmark performance is exactly what we need for the historical record, and the consolidated _clean_line() helper keeps the extraction logic incredibly clean. Great work driving this bug fix over the line! Approved and merging.

Issue #74, opened 2026-03-24
Refactor backend toward the v0.3 modular architecture


Kevin-TCUS:
# Refactor backend toward the v0.3 modular architecture

## 🔗 Relationship to prior work
This ticket is the structural backbone for the v0.3 pipeline. It enables the clean implementation of:
* **#69** Initialize and validate 4-tab Google Sheet schema
* **#70** Generate and standardize submission_id across the intake pipeline
* **#71** Implement append-only write logic for new submissions
* **#72** Upload resumes to Drive with deterministic naming
* **#73** Replace hardcoded Sheet column indices with repo-owned schema mapping

This refactor helps us move from a prototype toward a backend that can support resolver integration, dashboard APIs, and a future database without turning `main.py` into a giant control tower.

---

## 🎯 What we’re doing
We are refactoring the backend into clearer layers so contributors can work with less confusion and less risk.

### **The Target Architecture**
* **main.py:** App startup, route registration, and startup validation. This file should become thinner over time.
* **config.py:** Central source of truth for environment/config values (Sheet IDs, Drive folder IDs, OAuth settings, etc.).
* **sheet_schema.py:** The repo-owned contract for Google Sheet tab names and headers.
* **storage/:** The infrastructure layer. Google Sheets and Google Drive read/write logic.
* **services/:** The "Brain." Workflow and business logic (intake orchestration, parsing flow, resolver flow).
* **parser/:** Pure parsing and extraction logic only.
* **resolver/:** Normalization, alias-map knowledge, and resolver schemas.
* **benchmarks/:** Evaluation tools and quality check scripts.
* **api/routes/:** (Optional next step) Route files that can gradually move out of `main.py`.

### **Old to New Mapping**
| Current File/Logic | Target Direction |
| :--- | :--- |
| `main.py` | Remains, but thinned out to startup/routing. |
| `sheets_sync.py` | `storage/sheets_repo.py` |
| `drive_sync.py` | `storage/drive_repo.py` |
| `parser.py` | `parser/parser.py` |
| Background `_parse_and_update` | `services/parser_service.py` |
| Intake orchestration | `services/intake_service.py` |

---

## 🏗️ Future Backend Shape (v0.3 Target)
```text
backend/
├── main.py
├── config.py
├── sheet_schema.py
├── api/
│   └── routes/
│       ├── health.py
│       ├── intake.py
│       ├── dashboard.py
│       └── resumes.py
├── services/
│   ├── intake_service.py
[...]

PR #117, opened 2026-04-09
Extract config.py: centralize all environment variable reads

briannadelariva:
Move every os.getenv/os.environ call from main.py, sheets_sync.py, and drive_sync.py into a single backend/config.py module. Other modules now import config values directly, giving a single source of truth for env configuration.

### DIFF
--- a/backend/config.py
+++ b/backend/config.py
@@ -1,0 +1,22 @@
+import os
+from dotenv import load_dotenv
+
+load_dotenv()
+
+# ---- Google Sheets ----
+GOOGLE_SHEET_ID = os.getenv("GOOGLE_SHEET_ID")
+SHEET_NAME = os.getenv("SHEET_NAME", "applicantsInfo")
+GOOGLE_SERVICE_ACCOUNT_JSON = os.getenv("GOOGLE_SERVICE_ACCOUNT_JSON")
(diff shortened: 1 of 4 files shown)


Kevin-TCUS (commented):
This looks fantastic, Brianna! Thank you for jumping in. This feels like a very sensible first slice of the modular backend refactor and maps nicely to the config.py step in Issue #74.

Before I merge, quick check: were you able to boot the backend locally and hit the /health route after this change, just to confirm the new config imports don’t introduce any startup issues?

Also, one quick Git tip for next time: it’s usually best to open PRs from a dedicated feature branch rather than directly from your fork’s main. That makes it much easier to work on multiple issues without commits getting tangled.

If the local boot + /health check passed, I should be in good shape to merge this. Nice work.

briannadelariva:
@Kevin-TCUS  Yes, I did the local boot and /health route successfully after this change. No issues. I also touched base on this in my Linkedin message, but I am unable to create feature branches at this time. I would need to be added as a collaborator to make branches, which is why I created the fork. Please let me know if you'd like to send the invite and I will create a feature branch going forward. Thank you!

Kevin-TCUS (dismissed):
Thanks for confirming the local boot, Brianna! I went ahead and resolved a quick merge conflict that popped up from another PR we just merged, and everything looks great. Merging this in now!

PR #123, opened 2026-04-09
Refactor write_base_row() to build sheet rows for issue 99

jliu233:
## 📌 Summary

Refactors `write_base_row()` to build sheet rows from `backend/sheet_schema.py` instead of hardcoded column indices.

---

## 🔗 Related Issues

* Closes #99
* Builds on #96, #97 

---

## 🛠️ Changes

* Refactored `write_base_row()` in `sheets_sync.py` to remove all hardcoded column indices (e.g., `row[0]`, `row[10]`)
* Introduced schema-driven row construction using:

  * `build_row("submissions", data_dict)` from `sheet_schema.py`
* Replaced manual row assembly with a structured `row_data` dictionary aligned to `SUBMISSIONS_HEADERS`
* Updated append target to `submissions` tab

---

## ✅ Scope

* Only refactors how the base row is constructed
* Preserves all existing data mappings (name, email, timestamp, etc.)
* Keeps `main.py` and API behavior unchanged
* No changes to parser write logic (`update_resume_in_sheet`)

---

## 🚫 Non-goals

* No schema validation logic (issue #98 postponed)
* No changes to intake form fields or backend contract
* No file restructuring

---

## 🧪 Testing

* Verified that new submissions are successfully appended to the Google Sheet
* Confirmed all fields appear in the correct schema-defined columns
* Ensured no regression in `/api/upload` flow

### DIFF
--- a/backend/sheets_sync.py
+++ b/backend/sheets_sync.py
@@ -69,26 +69,22 @@ def write_base_row(
     ui_data: Optional[Dict[str, Union[str, List[str], bool]]] = None,
 ) -> None:
     """
-    Appends a base row using schema-driven row construction for the
-    submissions tab.
+    Appends a base row using schema-driven row construction.
+    This refactor changes only how the row is built, not the write-path behavior.
     """
-    if not submission_id:
-        raise ValueError("submission_id is required")
+    ts = _local_timestamp()
 
     if ui_data is None:
         ui_data = {}
 
-    ts = _local_timestamp()
-    drive_url = drive_file_url or _drive_link(drive_file_id)
-
     row_data = {
-        "submission_id": submission_id,
+        "submission_id": submission_id or "",
         "created_at": ts,
         "full_name": ui_data.get("name", ""),
         "email": ui_data.get("email", ""),
         "location_raw": ui_data.get("location", ""),
         "timezone": "",
-        "skills_raw": ui_data.get("areas", ""),
+        "skills_raw": "",
         "interests": ui_data.get("areas", ""),
         "experience_level": ui_data.get("experience", ""),
         "availability": ui_data.get("capacity", ""),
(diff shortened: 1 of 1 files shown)


Kevin-TCUS (changes requested):
Hi Jiaqi — this is very close, and the overall direction is right. I like the move to `build_row("submissions", row_data)` and the removal of the hardcoded row indices.

I do need a few changes before merge, because this ticket is meant to refactor *only* how the row is built, not change the write-path behavior itself.

A few things in the current diff go beyond #99:

1. **The append target changed from `f"{SHEET_NAME}!A2"` to `"submissions!A2"`.** That is a runtime behavior change, not just a row-construction refactor. For this ticket, please keep the existing target behavior unchanged.
2. **`resume_filename` is currently being set to `drive_url`.** That looks semantically incorrect — filename and URL are different fields. Please preserve the existing logical mapping here rather than introducing a new one in this PR.
3. **`skills_raw` and `interests` are both being populated from `ui_data.get("areas", "")`.** That introduces a new schema/data-model decision rather than just preserving the previous write behavior. For this ticket, please keep the mapping limited to preserving the existing logical values and do not invent new field semantics.
4. **The new `if not submission_id:` guard changes function behavior.** Since this ticket is scoped to row construction only, please avoid introducing new runtime requirements unless they are already required by the current flow.

Once those are tightened up, I think this will be in good shape to merge. The core schema-driven refactor is solid — I just want the PR to stay narrowly aligned with #99.

Kevin-TCUS (approved):
Changes look perfect, Jiaqi! I really appreciate the quick turnaround and your attention to keeping the scope tight. Approving and merging now.

When you are ready for your next slice, just pull main and you are clear to jump right into Issue #128 to apply this exact same schema-builder pattern to the parser writeback flow!

Issue #120, opened 2026-04-09
[Task] Refactor parser_results write path to use schema-driven row building


Kevin-TCUS:
### 🔗 Relationship to prior work
This is the next atomic implementation step after #99. It directly fulfills part of #73 (Replace hardcoded Sheet column indices) and continues the schema-driven transition started with `backend/sheet_schema.py` and its helper functions.

### 🎯 Summary
Right now, the parser writeback flow constructs the `parser_results` row using a hardcoded array and implicit field ordering. This is fragile and makes the code harder to maintain as the schema evolves.

We are replacing that with schema-driven row construction. The code should use the field names defined in `backend/sheet_schema.py` and `build_row("parser_results", data_dict)` to create the final row automatically.

*This ticket is intentionally narrow:* Refactor the construction logic of the `parser_results` write path only.

### 🛠️ Implementation Steps
1. **Update the parser writeback logic to use Schema Helpers:** Locate the current logic that assembles the `parser_results` row for Google Sheets.
2. Instead of building a raw ordered list manually, **create a dictionary** where keys match the `PARSER_RESULTS_HEADERS` field names (for example: `submission_id`, `parser_run_id`, `parsed_skills_raw`, `resolved_skill_ids`).
3. **Call `build_row("parser_results", data_dict)`** to generate the final list for the Google Sheets API.
4. **Maintain Data Integrity:** The goal is to change *how* the row is built, not *what* is being written. Ensure the same logical data is still written to the `parser_results` tab, just via schema-driven mapping instead of hardcoded ordering.
5. **Improve Readability:** Remove manual array construction and implicit positional assumptions. The resulting code should make it obvious which parser/resolver outputs map to which schema fields.

### ✅ Definition of Done
- [ ] The `parser_results` write path no longer contains hardcoded ordered row arrays for this tab.
- [ ] Row construction is exclusively driven by `build_row()` from `backend/sheet_schema.py`.
- [ ] The backend successfully writes parser results to the `parser_results` tab with data in the correct columns.
- [ ] *No unrelated refactors:* Do not move files or refactor other write/update flows yet.

### 🚫 Non-goals
- Refactoring startup schema validation.
- Moving files into `storage/` or `services/` layers.
- Changing parser or resolver behavior.
- Changing the `parser_results` schema itself.

### 📝 One-Sentence PR Rule
Your PR description for this task should simply be:
[...]

Issue #70, opened 2026-03-24
Generate and standardize submission_id across the intake pipeline


Kevin-TCUS:
## 🔗 Relationship to prior work
This ticket builds directly on top of the work in **#40**. 
* **#40** focused on removing the in-memory counter and using `submission_id` in the current upload flow.
* **This ticket** expands that idea for v0.3. We want `submission_id` to become the **canonical linking key** that connects the full Libelle pipeline.

*Note: If some of this is already completed in #40, that work should be treated as partial progress toward this goal.*

---

## 🎯 What we’re doing
We need one stable ID (`submission_id`) that follows a volunteer’s application through every stage of the system—from the moment they hit "Submit" to storage, parsing, resolver output, and finally the dashboard review. 

## 🛠️ What you need to do
- [ ] **Generate the ID:** Ensure the backend uses `uuid.uuid4()` at the very start of a new submission handler.
- [ ] **Write it to the Sheet:** Ensure the ID is stored in the `submissions` tab (targeting Column A as defined in the new schema).
- [ ] **Use it in Drive filenames:** Use this format for uploaded resumes: `{submission_id}_{original_filename}`. This allows reviewers to see the volunteer's name while maintaining a unique ID.
- [ ] **Pass it through the pipeline:** Ensure the same ID is passed into the parser and resolver execution contexts so their outputs can be linked back to the original submission.
- [ ] **Return it in the API response:** Ensure the API sends the `submission_id` back in the success response.
- [ ] **Use it in logs/errors:** Update backend logs so a single submission can be traced end-to-end using this ID.

## 💡 Why this matters
* **Traceability:** If a submission fails halfway through, we can use the ID to find exactly where it got stuck.
* **Reviewer UX:** Reviewers can see both a unique ID and a human-readable name in Google Drive.
* **Database Readiness:** This establishes the "Primary Key" pattern we will need when we eventually move to a real database like Postgres.

## ✅ How to tell it’s done
1. You submit a volunteer application.
2. You see one UUIDv4 in the API response.
3. That exact same ID appears in the `submissions` record.
4. That same ID is prefixed to the resume filename in Drive.
5. Backend logs show the entire "journey" of that specific ID.

## 🏷️ Labels
`feat`, `backend`, `epic:pipeline-hardening`, `api`

## 🏁 Milestone
`v0.3 – Trustworthy Volunteer Intake`

Kevin-TCUS:
#129 completed via #205. Next step is #126.

[...]

PR #146, opened 2026-04-17
Extract services/ layer: intake and parser orchestration (#74)

briannadelariva:
## Summary
- New `backend/services/` package:
  - `intake_service.py` exposes `validate_intake` (field + file-type validation) and `finalize_submission` (size check, text extraction, `submission_id` generation, Drive upload + Sheets append).
  - `parser_service.py` owns the background `parse_and_update` workflow.
- `main.py` is now startup + HTTP glue only: FastAPI app, CORS, exception handlers, and thin route handlers that call the services. **337 → 190 lines.**
- Services own a plain-Python `IntakeError`; the route layer translates it to `HTTPException` so services stay free of FastAPI concerns.

Implements the `services/` extraction step from #74. Sets up the clean boundary the UUIDv4 migration (#70) and schema-driven write work (#98, #99, #120) will land on top of.

## No behavior changes
Error codes, payload shapes, and validation ordering are preserved. The split entry points also let the route call `validate_intake` before `await file.read()`, so PDF bytes are no longer loaded into memory on 422 `VALIDATION_ERROR` or 400 `INVALID_FILE_TYPE` rejections — a small efficiency win with no user-visible change.

Verified end-to-end with FastAPI `TestClient`:

- `GET /health` → 200 ok
- `POST /api/upload` (no file) → 400 `FILE_REQUIRED`
- `POST /api/upload` (bad fields) → 422 `VALIDATION_ERROR` with per-field map
- `POST /api/upload` (non-PDF) → 400 `INVALID_FILE_TYPE`

## Test plan
- [ ] CI passes
- [ ] Reviewer verifies `main.py` route handlers contain no business logic beyond request parse → service call → response format
- [ ] Reviewer runs backend locally and exercises a valid intake submission end-to-end (hits Drive + Sheets — not exercisable in the PR author's environment)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -1,29 +1,20 @@
-from fastapi import FastAPI, File, UploadFile, BackgroundTasks, HTTPException, Form, Request
-from fastapi.responses import JSONResponse
-from fastapi.middleware.cors import CORSMiddleware
-from fastapi.exceptions import RequestValidationError
-from typing import Optional, Dict, Any, List, Union
-import fitz
 import traceback
-import json
-import uuid
-import re
 from datetime import datetime, timezone
+from typing import Any, Dict, Optional
 
-from parser import parse_resume
-from storage.sheets_repo import write_base_row, update_resume_in_sheet
-from storage.drive_repo import upload_pdf, build_auth_url, exchange_code
-from config import (
-    MAX_PDF_MB, ALLOWED_ORIGINS, APP_REDIRECT_URI,
-)
+from fastapi import BackgroundTasks, FastAPI, File, Form, HTTPException, Request, UploadFile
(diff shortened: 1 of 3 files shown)


Kevin-TCUS (approved):
This is a really strong execution of the services-layer extraction. I especially like the IntakeError pattern and keeping HTTP translation in main.py rather than leaking FastAPI concerns into the service layer. The split between validate_intake and finalize_submission is also a nice improvement, since invalid requests now fail before loading PDF bytes into memory.This cleanly advances #74 and gives the backend a much better boundary for the next round of UUID / schema / parser write work. Approved and merging.

PR #137, opened 2026-04-16
Add pure schema comparison helpers for tabs and headers for issue 127

jliu233:
## 🎯 Summary

Adds pure schema comparison helpers for tabs and headers in `backend/sheet_schema.py` to support later startup validation.

---

## 🔗 Related Issue

Closes #127
Part of #69 (Initialize and validate 4-tab Google Sheet schema)

---

### 1. `compare_headers(tab_name, actual_headers)`

* Compares actual sheet headers against schema-defined headers
* Returns structured mismatch details:

  * `is_match`
  * `missing_expected`
  * `extra_found`
* Uses `get_headers(tab_name)` (no duplication of schema logic)

### 2. `compare_tab_names(actual_tabs)`

* Dynamically derives expected tabs from `SHEET_SCHEMA.keys()`
* Compares against provided tab list
* Returns:

  * `is_match`
  * `missing_expected`
  * `extra_found`

---

## ✅ Design Notes

* Pure helper logic only (no side effects)
* No Google Sheets API calls
* No startup wiring or runtime validation yet
* Fully reusable for future validation layer (#98)

---

## 🚫 Non-Goals

* No changes to `main.py`
* No schema enforcement or exceptions
* No sheet reads/writes
* No tab creation/modification

---

## 🧪 Testing

* Manually verified expected vs actual comparisons:

  * Missing headers
  * Extra headers
  * Exact match cases
  * Missing / extra tabs

---

## 📌 Next Step

These helpers will be used in #98 to implement startup schema validation.

PR #150, opened 2026-04-18
Add startup schema validation for issue #98

briannadelariva:
Adds startup validation so the backend checks the live Google Sheet against `backend/sheet_schema.py`.

Closes #98.

## What changed

- `backend/validator.py` (new) — `validate_sheet_schema()` orchestrates the comparison; raises `SchemaValidationError` on any mismatch.
- `backend/storage/sheets_repo.py` — new `fetch_live_schema()` helper: one `spreadsheets().get(fields="sheets.properties.title")` + one `values().batchGet` for Row 1 of every expected tab that exists.
- `backend/main.py` — second `@app.on_event("startup")` handler that calls `validate_sheet_schema()`. Runs after the existing `_startup_log`, so `[STARTUP]` lines print first and `[SCHEMA] Schema Validated` (or the failure traceback) appears right after.
- `backend/tests/test_validator.py` — four `monkeypatch`-based unit tests covering happy path, header mismatch, missing tab, and empty Row 1.

`backend/sheet_schema.py` is the only source of truth; the validator imports `SHEET_SCHEMA` and the pure comparison helpers (`compare_tab_names`, `compare_headers`) from #137. No headers are re-listed anywhere.

## Behavior

**Success** — prints:
```
[SCHEMA] Schema Validated
```

**Failure** — raises `SchemaValidationError` with an actionable message naming every offending tab. Example (header rename + missing tab):
```
Google Sheet schema validation failed:

[tabs] Sheet tab set does not match SHEET_SCHEMA.
  expected: ['submissions', 'parser_results', 'ops', 'errors']
  actual:   ['submissions', 'parser_results', 'ops']
  missing tabs: ['errors']

[submissions] Header mismatch.
  expected: [..., 'email', ...]
  actual:   [..., 'email_address', ...]
  missing:  ['email']
  extra:    ['email_address']
```

Uvicorn treats the uncaught exception from the startup handler as a fatal boot failure (`Application startup failed. Exiting.`), so the service never begins accepting requests on a bad schema — no silent continue.

## Non-goals (per #98)

- No auto-creation of tabs or repair of headers.
- No changes to `sheet_schema.py`, `config.py`, or existing write-path code.
- No `SKIP_SCHEMA_VALIDATION` escape hatch.

## ⚠️ Prerequisite before deploying against the live sheet

The live Google Sheet still has the legacy single-tab structure (`applicantsInfo`) and has **not** been migrated to the 4-tab schema this validator expects. Verified locally — uvicorn currently halts at startup with:

```
[tabs] Sheet tab set does not match SHEET_SCHEMA.
  expected: ['submissions', 'parser_results', 'ops', 'errors']
  actual:   ['applicantsInfo']
  missing tabs: ['submissions', 'parser_results', 'ops', 'errors']
  extra tabs:   ['applicantsInfo']
```

Before deploying this change, a sheet admin must:
1. Create the four tabs: `submissions`, `parser_results`, `ops`, `errors`.
2. Populate Row 1 of each with the exact header order from `backend/sheet_schema.py`.
3. Decide what to do with the existing `applicantsInfo` rows (migrate columns into `submissions`, archive, or delete) — the validator rejects extra tabs, so it can't coexist.

This is tracked as a manual admin step per #98's non-goals. The PR ships the enforcement layer; the sheet migration is a separate task.

## Test plan

- [x] `pytest backend/tests -q` — 4/4 pass (happy path, header mismatch, missing tab, empty Row 1).
- [x] `python -c "import main"` — both startup handlers registered in order: `['_startup_log', '_startup_validate_schema']`.
- [x] `uvicorn main:app` locally against the current live sheet — validation correctly fails with the expected error format and the server refuses to start.
- [ ] After the sheet is migrated, re-run `uvicorn main:app` and confirm `[SCHEMA] Schema Validated` appears and the server accepts requests on `/health`.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -12,7 +12,6 @@ DRIVE_ROOT_FOLDER_ID=PASTE_DRIVE_FOLDER_ID
 
 # Google Sheets destination
 GOOGLE_SHEET_ID=PASTE_SHEET_ID
-SHEET_NAME=applicantsInfo
 
 # Sheets credentials (choose ONE)
 
--- a/backend/config.py
+++ b/backend/config.py
@@ -5,7 +5,6 @@ load_dotenv()
 
 # ---- Google Sheets ----
 GOOGLE_SHEET_ID = os.getenv("GOOGLE_SHEET_ID")
-SHEET_NAME = os.getenv("SHEET_NAME", "applicantsInfo")
 GOOGLE_SERVICE_ACCOUNT_JSON = os.getenv("GOOGLE_SERVICE_ACCOUNT_JSON")
 GOOGLE_CREDENTIALS = os.getenv("GOOGLE_CREDENTIALS", "org_credentials.json")
 
--- a/backend/main.py
+++ b/backend/main.py
@@ -11,6 +11,7 @@ from config import ALLOWED_ORIGINS, APP_REDIRECT_URI, MAX_PDF_MB
 from services.intake_service import IntakeError, finalize_submission, validate_intake
 from services.parser_service import parse_and_update
 from storage.drive_repo import build_auth_url, exchange_code
+from validator import validate_sheet_schema
 
 
 app = FastAPI(title="Libelle Backend API")
@@ -31,6 +32,11 @@ def _startup_log():
     print("[STARTUP] Libelle backend booted")
     print(f"[STARTUP] MAX_PDF_MB={MAX_PDF_MB}")
     print(f"[STARTUP] ALLOWED_ORIGINS={ALLOWED_ORIGINS}")
+
+
+@app.on_event("startup")
+def _startup_validate_schema():
+    validate_sheet_schema()
 
 
 # -----------------------------
--- a/backend/storage/sheets_repo.py
+++ b/backend/storage/sheets_repo.py
@@ -5,13 +5,14 @@ from typing import Optional, Dict, Union, Any, List
 from datetime import datetime, timezone
 from googleapiclient.discovery import build
 
-from config import GOOGLE_SHEET_ID, SHEET_NAME
+from config import GOOGLE_SHEET_ID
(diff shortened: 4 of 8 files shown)


Kevin-TCUS (commented):
This looks incredibly strong, @briannadelariva . The validator structure, startup wiring, and failure messaging are exactly what we need for #98.

I did catch one small legacy ghost in sheets_repo.py, though: write_base_row() is still relying on SHEET_NAME from config. Since v0.3 now strictly contracts that tab as submissions, this creates a runtime mismatch risk where startup validation can pass but the first upload still tries to write to an old tab name from .env.

Could you make one small follow-up change here before merge:
- add a SUBMISSIONS_SHEET_NAME = "submissions" constant, similar to PARSER_RESULTS_SHEET_NAME
- use that in the submissions write path instead of SHEET_NAME

If you have time, I’d also love one quick local sanity run against a copy of the new 4-tab v0.3 template to confirm:
- [SCHEMA] Schema Validated
- /health works
- and ideally one happy-path upload succeeds

Once that’s in, I’m very comfortable merging.

briannadelariva:
Done — pushed a follow-up commit (0634723) on this PR that addresses the review: added a `SUBMISSIONS_SHEET_NAME = "submissions"` constant, swapped `write_base_row()` over to it, and dropped the now-dead `SHEET_NAME` from `config.py`, `.env.example`, and the local-dev doc. Ready for another look.

Kevin-TCUS (approved):
This is absolutely perfect, Brianna. Thank you for not only making the SUBMISSIONS_SHEET_NAME swap, but also for going the extra mile to remove the dead SHEET_NAME references from the environment templates and local docs. That is excellent cleanup and keeps the write path aligned with the new v0.3 contract. Approved and merging.

PR #152, opened 2026-04-18
Add GitHub Actions PR workflow for backend tests

### LINKED ISSUE #151: [CI] Add GitHub Actions PR workflow for backend tests
Sub-item of #115 (DX / Reliability Layer for v0.3).

### Goal
Add a lightweight GitHub Actions workflow that runs `pytest backend/tests` on every PR touching `backend/**` so backend regressions are caught before merge instead of at runtime.

### Scope
- New `.github/workflows/backend.yml`
- New `backend/requirements-dev.txt` with `pytest` pinned
- Path-filtered trigger so frontend-only PRs don't pay the Python setup cost
- Pip cache via `actions/setup-python` with `cache: 'pip'` for speed

### Explicitly out of scope (per #115)
- Lint / typecheck enforcement
- Frontend build check (separate sub-item in #115)
- Matrix testing across Python versions
- Docker / Compose
- Coverage / JUnit reporting

### Done when
- Workflow runs automatically on any PR touching `backend/**`
- A failing backend test blocks the PR check
- A frontend-only PR does not trigger the backend workflow

briannadelariva:
Closes #151.
Sub-item of #115 (DX / Reliability Layer for v0.3).

## Summary
Adds a minimal GitHub Actions workflow that runs `pytest backend/tests -q` on every PR touching `backend/**` (or the workflow file itself). Dev-only test deps live in a new `backend/requirements-dev.txt` so runtime `requirements.txt` stays lean.

## What changed
- `.github/workflows/backend.yml` (new) — Ubuntu-latest, Python 3.11, path-filtered on `backend/**` + the workflow file, pip cache keyed on both requirements files.
- `backend/requirements-dev.txt` (new) — `pytest==9.0.3`. Contributors who want to run tests locally add `pip install -r backend/requirements-dev.txt` to their setup.

## Scope
Deliberately tight, matching #115's "lightweight DX layer" framing. Explicitly **not** in this PR:
- Lint / typecheck enforcement
- Frontend build check (separate sub-item in #115)
- Matrix testing across Python versions
- Coverage / JUnit reporting
- Docker / Compose standardization

Those can ship as their own atomic PRs when their sub-items get picked up.

## Test plan
- [x] `pytest backend/tests -q` passes locally (4/4, same as #150's baseline).
- [x] `pip install -r requirements.txt -r requirements-dev.txt` resolves cleanly in a fresh venv.
- [ ] This PR itself exercises the workflow — the `backend/requirements-dev.txt` and workflow-file changes both match the path filter, so the new CI check will run on this PR and should go green.
- [ ] A follow-up frontend-only PR should *not* trigger this workflow (verifies path filter is doing its job).


Kevin-TCUS (approved):
This is a great addition. Lightweight, practical, and exactly the kind of reliability layer we should be adding at this stage. I like that the workflow is path-filtered so backend tests only run when backend files change, and keeping pytest in requirements-dev.txt is a clean separation from runtime dependencies. Approved and merging.

Issue #149, opened 2026-04-17
[Parser] Targeted extraction fixes from benchmark analysis: reading order, header spacing, and skills sub-header flattening


Kevin-TCUS:
### Relationship to prior work
This issue is derived directly from the recent benchmark analysis artifact identifying the main causes of low parser performance across several resumes. The goal is to turn that analysis into a bounded parser-improvement ticket that can be tested against the benchmark harness.

### Summary
Recent benchmark review identified three concrete failure modes contributing to low skills/location extraction performance:
1. PDF reading-order scrambling
2. Garbled intra-word character spacing affecting header detection
3. Sub-header flattening issues inside skills sections

This issue is intentionally framed as one targeted parser-improvement lane so the analysis can be converted into executable code changes and benchmarked as a unit. The implementation does not need to solve every edge case perfectly, but it should make measurable progress against the known failure patterns.

### Observed Issues

**1. PDF reading-order scrambling**
PyMuPDF text extraction can return text in internal PDF storage order rather than visual reading order. In some resumes, this causes section headers and body content to appear out of sequence, especially in multi-column or visually complex documents.
* *Examples observed:* `low_signal` resumes scoring 0 on skills F1; `embed_link_02` complete failure on skills.

**2. Garbled character spacing from font encoding**
Some PDFs extract with spurious intra-word spacing in likely header tokens, which causes section headers to be missed by the parser.
* *Example pattern:* `"S k i l l s"` instead of `"Skills"`.

**3. Skills sub-header flattening**
Some resumes organize skills under labeled sub-headings such as:
* `Programming: Python, R, SQL`
* `Tools: Git, Excel`

The current parser preserves these labels rather than flattening the actual skill values, which produces false negatives against the benchmark gold JSON.

### Scope
Contributor should implement targeted fixes against one or more of the above failure modes while keeping the PR reviewable and benchmark-driven.

**Suggested implementation direction:**
* Improve extraction ordering for affected resumes where feasible.
* Add lightweight cleanup for likely header token spacing artifacts before section parsing runs.
* Flatten labeled sub-headers inside the skills section so only the skill values are preserved in the final output.

### Definition of Done
* [ ] At least one of the identified failure modes is improved in code.
* [ ] Changes remain tightly scoped to parser/extraction cleanup and do not redesign the parser broadly.
* [ ] Benchmark is rerun against the affected cases.
* [ ] PR description includes before/after notes referencing the resumes or scenario types improved.
[...]

Issue #84, opened 2026-03-29
[Benchmark] Add saved aggregate statistical summary artifact for parser runs


Kevin-TCUS:
# [Benchmark] Add saved aggregate statistical summary artifact for parser runs

## Relationship to prior work
This issue builds directly on the existing benchmark harness in `scripts/benchmark.py`, built according to [this issue](https://github.com/The-Chamber-of-Us/libelle/issues/48).

Currently, the harness:
* Scores row-level parser results.
* Computes TP / FP / FN per resume, parser, and field.
* Computes precision / recall / F1 per row.
* Prints a quick aggregate summary to the CLI at the end of the run.

**The Gap:** It does not yet save that aggregate summary anywhere as a durable artifact. Once you close your terminal, that high-level data is gone unless you manually scroll up and copy it.

## The Mission
Turn the current end-of-run CLI summary into a real **benchmark summary artifact** (`summary.md`) that is saved automatically alongside the other benchmark outputs in the `runs/<timestamp>/` folder.

This is a blend of **software engineering** (decoupling logic) and **evaluation science** (statistical rigor).

## Why this matters
In fields like **Biomed** or **Applied ML**, it is not enough to know if a system performs "well on average." You need to understand the distribution of that performance. 

By adding a saved summary, we can answer:
1. **Is the parser actually improving?** (By comparing this run's `summary.json` to a previous one).
2. **Is performance stable?** (By comparing Micro vs. Macro averages).
3. **Where are the hallucinations?** (By identifying failure concentrations).

### 1. Save Aggregate Parser/Field Metrics
For each parser and field combination, compute and save:
* Total TP, FP, and FN.
* **Micro-Precision, Micro-Recall, and Micro-F1.**
  * *Note:* Micro-averaging aggregates all individual TPs/FPs/FNs across all resumes first, then calculates the score.

### 2. Add Macro-Level Summary (The "Stats Flex")
Compute and save **Macro-F1** for each parser/field pair.
* **Macro-F1** is the arithmetic mean of the individual F1 scores calculated for each resume.
* *Math check:* $$F_1 = 2 \cdot \frac{\text{Precision} \cdot \text{Recall}}{\text{Precision} + \text{Recall}}$$
* Comparing Micro vs. Macro helps us see if the parser is failing specifically on "large" resumes (many skills) or "small" ones.

### 3. Generate `summary.md`
Produce a beautifully formatted Markdown table in the run folder. It should include:
[...]

Issue #82, opened 2026-03-27
Implement Field-Level Scoring in Benchmark Harness


Kevin-TCUS:
# Implement Field-Level Scoring in Benchmark Harness

## 🔗 Relationship to prior work
This is a logic enhancement to the benchmark harness currently being finalized in **#77**.

## 🎯 The Upgrade
Right now the thinking is: **“Does the parsed result match the sample resume?”** We need to make it more structured to gain better insight into where the parser is succeeding and where it is failing.

---

## 🛠️ What you need to do
Move from a "whole resume score" to individual **field-level scoring functions**.

### 1) Implement Functional Scorers
Add specific functions for each field. Each returns a float between **0.0 and 1.0**:
* `score_name()`
* `score_email()`
* `score_location()`
* `score_skills()`

### 2) Use Fuzzy Logic (Example)
Instead of a binary True/False, use nuanced returns to handle "near misses":

```python
def score_email(pred, truth):
    if pred == truth:
        return 1.0
    elif pred and truth and pred.lower() == truth.lower():
        return 0.9  # Case mismatch
    elif pred:
        return 0.3  # Found something, but it's wrong
    else:
        return 0.0  # Missing entirely
```
### 3) Track "Why" (The Reason Layer)
For the next step, track why something scored low. This will be essential for debugging, gathering training data, and improving the Resolver V1 later.

**Example structured output:**
```json
{
  "field": "skills",
  "score": 0.47,
  "notes": "missed 3 expected skills, 2 fuzzy matches"
}
```
### 💡 Why this matters
* **Nuance:** We get credit for "near misses" rather than flat zeros.
* **Debuggability:** We can identify if a specific field (like Location) is failing across the entire dataset.
* **Metric Integrity:** This provides the "Ground Truth" needed to safely transition from Regex to AI-augmented parsing.

### ✅ How to tell it’s done
* The benchmark script output now includes a per-field breakdown.
* The total benchmark score is a weighted average of these field scores.
* Scores below **1.0** include a `notes` or `reason` entry in the logs.

**🏷️ Labels:** `enhancement`, `benchmarks`, `math-heavy`

Kevin-TCUS:
Closing as superseded by the current benchmark harness direction.

This issue was opened when the benchmark concept was still early and proposed fuzzy field-level scoring across name, email, location, and skills.

The v0.3 benchmark direction has since become more specific and measurable:

- focus on skills and location first
- report TP / FP / FN examples
[...]

PR #154, opened 2026-04-18
Fix async parser writeback: plumb submission_id through parse_and_update

### LINKED ISSUE #153: [Bug] Async parser never writes parser_results — update_resume_in_sheet() signature mismatch
Surfaced while running the local sanity check for PR #150 against a freshly-provisioned 4-tab sheet.

`POST /api/upload` succeeds end-to-end: PDF lands in Drive, base row appends to `submissions`. The background parser task then crashes before it can write anything to `parser_results`.

### Traceback
```
File "backend/services/parser_service.py", line 14, in parse_and_update
    update_resume_in_sheet(parsed)
TypeError: update_resume_in_sheet() missing 1 required positional argument: 'parsed'
```

### Root cause
Contract drift between call site and callee:

- `backend/services/parser_service.py:14` — `update_resume_in_sheet(parsed)` (1 arg)
- `backend/storage/sheets_repo.py:178` — `def update_resume_in_sheet(submission_id: str, parsed: Dict[str, Any])` (2 args, `submission_id` required)

The caller chain also never plumbs `submission_id` through — `main.py:174` adds the background task as `parse_and_update(drive_file_id, pre_text)` with no `submission_id`, and `parser_service.parse_and_update` doesn't accept one.

### Impact
- `parser_results` tab is never populated for any submission.
- `errors` tab is also never populated (parser_service prints + `traceback.print_exc()` only).
- Sync path (Drive upload + `submissions` row) is unaffected.

### Suggested fix
1. `main.py:174` — pass `result["submission_id"]` into the background task.
2. `parser_service.parse_and_update` — accept `submission_id: str` as the first arg.
3. `parser_service.py:14` — `update_resume_in_sheet(submission_id, parsed)`.

### Test coverage
Add a `monkeypatch`-based regression test for `parse_and_update` that stubs `parse_resume` + `update_resume_in_sheet` and asserts the callee receives `(submission_id, parsed)`. Slots directly under the "[Test] Add parser writeback regression test" sub-item of #115.

This is exactly the "signature mismatch only discovered at runtime" pattern #115 names in its *Why this matters* section. The backend CI workflow from #152 will enforce the new test on future PRs.

briannadelariva:
Closes #153.
Related to #115 (DX / Reliability Layer for v0.3).

## Summary
The background parser task was crashing before it could write anything to the `parser_results` tab. Every submission's sync path (Drive upload + `submissions` row) succeeded, but the async writeback silently failed. Surfaced while running the local sanity verification Kevin requested for #150.

## Root cause
`services/parser_service.parse_and_update` called `update_resume_in_sheet(parsed)` with 1 argument, but `storage/sheets_repo.update_resume_in_sheet` expects `(submission_id: str, parsed: Dict[str, Any])`. The caller chain (`main.py:174`) also never plumbed `submission_id` through the background task.

## What changed
- `backend/main.py` — pass `result["submission_id"]` into the background task.
- `backend/services/parser_service.py` — accept `submission_id: str` as the first arg and forward it to `update_resume_in_sheet`.
- `backend/tests/test_parser_service.py` (new) — monkeypatched regression covering:
  - the fixed `(submission_id, parsed)` contract with `drive_file_id` still stitched into `parsed`
  - the existing fire-and-forget behavior where `parse_and_update` swallows exceptions so the background task doesn't propagate

## Test plan
- [x] `pytest backend/tests -q` — 6/6 pass (4 validator + 2 new parser service).
- [x] Live end-to-end: `uvicorn main:app`, then `POST /api/upload` with a real PDF. `parser_results` tab now receives a row (previously always empty).
- [x] Backend CI workflow from #152 will run the new test on this PR.

## Note on scope
This is a minimal fix for the signature drift + a regression test that slots directly into #115's *[Test] Add parser writeback regression test* sub-item. Broader parser-coverage tests (extraction quality, resolver integration) are Riya's benchmark workstream (#149, #84, #82) and intentionally out of scope here.

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -171,7 +171,12 @@ async def upload_volunteer_application(
             },
         )
 
-    background_tasks.add_task(parse_and_update, result["drive_file_id"], result["pre_text"])
+    background_tasks.add_task(
+        parse_and_update,
+        result["submission_id"],
+        result["drive_file_id"],
+        result["pre_text"],
+    )
 
     return JSONResponse(
         status_code=200,
--- a/backend/services/parser_service.py
+++ b/backend/services/parser_service.py
@@ -5,14 +5,14 @@ from parser import parse_resume
 from storage.sheets_repo import update_resume_in_sheet
 
 
-def parse_and_update(drive_file_id: str, pre_extracted_text: str = "") -> None:
+def parse_and_update(submission_id: str, drive_file_id: str, pre_extracted_text: str = "") -> None:
     """Parse the extracted resume text and update the parser_results row in Sheets."""
     try:
-        print(f"[JOB] Parsing drive_file_id={drive_file_id} ...")
+        print(f"[JOB] Parsing submission_id={submission_id} drive_file_id={drive_file_id} ...")
         parsed = parse_resume(pre_extracted_text or "")
         parsed["drive_file_id"] = drive_file_id
-        update_resume_in_sheet(parsed)
-        print(f"[JOB] Parsed + updated sheet drive_file_id={drive_file_id}")
+        update_resume_in_sheet(submission_id, parsed)
+        print(f"[JOB] Parsed + updated sheet submission_id={submission_id}")
     except Exception as e:
-        print(f"[JOB] Error parsing drive_file_id={drive_file_id}: {e}")
+        print(f"[JOB] Error parsing submission_id={submission_id}: {e}")
         traceback.print_exc()
--- a/backend/tests/test_parser_service.py
+++ b/backend/tests/test_parser_service.py
@@ -1,0 +1,41 @@
+from services import parser_service
+
+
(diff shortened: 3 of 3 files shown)


Kevin-TCUS (approved):
Great catch on the [signature mismatch](https://github.com/The-Chamber-of-Us/libelle/issues/153). This fix for the submission_id plumbing looks solid and solves the silent failure we saw during the sanity run.

I really appreciate the [regression test](https://github.com/The-Chamber-of-Us/libelle/pull/154/files#diff-8c6523998f828a1c97793910f545464109400030573760447155681146313364)—it makes it very clear how the data is supposed to flow and ensures we don't break this again in future updates. Keeping the 'fire-and-forget' logic for the async path is the right call for stability.

Approved and merging.

PR #159, opened 2026-04-21
added summary.md, summary.json and aggregated stats

### LINKED ISSUE #84: [Benchmark] Add saved aggregate statistical summary artifact for parser runs

jriya186:
**Updated description**
## Summary
Adds `summary.md` and `summary.json` as saved artifacts to each benchmark run, alongside the existing `report.csv`, `examples.md`, and `run_log.json`. Fixes #84 

## Changes
- Added `compute_aggregate_stats()` — extracts aggregation logic from `_print_summary()` into a standalone reusable function that computes micro and macro metrics; called once in `main()` and passed into all downstream functions
- Added `write_summary_md()` — generates a formatted markdown summary including run metadata, a scoreboard table, and a failure heatmap
- Added `write_summary_json()` — saves aggregate stats as JSON using unrounded raw values for macro_f1 and std_dev, for future automated regression testing
- Refactored `_print_summary()` to accept the pre-computed stats object instead of recomputing internally

## Metrics included
- Micro-F1 (pools all TP/FP/FN across resumes, then computes)
- Macro-F1 (mean of per-resume F1 scores)
- Standard deviation of per-resume F1 scores
- Top 3 resume/field combinations by false positive count (failure heatmap)

### DIFF
--- a/scripts/benchmark.py
+++ b/scripts/benchmark.py
@@ -333,6 +333,71 @@ def write_run_log(
         json.dump(log, f, indent=2)
     return path
 
+# Summary markdown file 
+def write_summary_md(rows: List[Dict], out_dir: Path, args: argparse.Namespace) -> Path:
+    stats = compute_aggregate_stats(rows)
+    path = out_dir / "summary.md"
+
+    # metadata
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    git_commit = _get_git_commit()
+    parsers_used = ", ".join(args.parsers)
+
+    # heatmap with top 3 resumes by FP count
+    top_fps = sorted(rows, key=lambda r: r["fp_count"], reverse=True)[:3]
+
+    lines = []
+
+    # writing metadata
+    lines.append("# Benchmark Summary\n")
+    lines.append(f"**Timestamp:** {timestamp}  ")
+    lines.append(f"**Git Commit:** `{git_commit}`  ")
+    lines.append(f"**Parsers:** {parsers_used}\n")
+
+    # writing scoreboard
+    lines.append("## Scoreboard\n")
+    lines.append("| Parser | Field | Micro-F1 | Macro-F1 | Std Dev |")
+    lines.append("|--------|-------|----------|----------|---------|")
+    for (parser, field), s in sorted(stats.items()):
+        lines.append(
+            f"| {parser} | {field} | {s['micro_f1']:.3f} | {s['macro_f1']:.3f} | {s['std_dev']:.3f} |"
+        )
+
+    # writing failure heatmap
+    lines.append("\n## Failure Heatmap (Top 3 by False Positives)\n")
+    lines.append("| Resume | Field | Parser | FP Count |")
+    lines.append("|--------|-------|--------|----------|")
+    for r in top_fps:
+        lines.append(f"| {r['resume']} | {r['field']} | {r['parser']} | {r['fp_count']} |")
+
+    with open(path, "w") as f:
+        f.write("\n".join(lines))
+
+    return path
+
+# Summary JSON file
+def write_summary_json(rows: List[Dict], out_dir: Path, args: argparse.Namespace) -> Path:
+    stats = compute_aggregate_stats(rows)
+    path = out_dir / "summary.json"
+
+    serializable = {}
+    for (parser, field), s in stats.items():
+        key = f"{parser}::{field}"
+        serializable[key] = {
+            "micro_p": s["micro_p"],
+            "micro_r": s["micro_r"],
+            "micro_f1": s["micro_f1"],
+            "macro_f1": s["macro_f1"],
+            "std_dev": s["std_dev"],
+            "tp": s["tp"],
+            "fp": s["fp"],
+            "fn": s["fn"],
+        }
+
+    with open(path, "w") as f:
+        json.dump(serializable, f, indent=2)
+
+    return path
 
 # ---------------------------------------------------------------------------
 # Main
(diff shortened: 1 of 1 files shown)


scripts/benchmark.py
```diff
@@ -466,35 +531,62 @@ def main():
     csv_path = write_report_csv(report_rows, out_dir)
     md_path = write_examples_md(report_rows, out_dir)
     log_path = write_run_log(out_dir, args.parsers, args, errors)
+    summary_path = write_summary_md(report_rows, out_dir, args)
```
Kevin-TCUS: Since we’re using the same aggregate stats across the Markdown, JSON, and CLI outputs, let’s compute stats = compute_aggregate_stats(report_rows) once right before this block and pass the shared object down into those functions instead of recalculating it three times.

scripts/benchmark.py
```diff
@@ -466,35 +531,62 @@ def main():
...
+        groups[key].append(r)
+
+    stats = {}
+    for (parser, field), group_rows in groups.items():
+        tp = sum(r["tp_count"] for r in group_rows)
+        fp = sum(r["fp_count"] for r in group_rows)
+        fn = sum(r["fn_count"] for r in group_rows)
+        micro_p, micro_r, micro_f1 = _compute_prf(tp, fp, fn)
+
+        f1_scores = [r["f1"] for r in group_rows]
```
Kevin-TCUS: One quick adjustment for the JSON artifact: r["f1"] was already rounded to 3 decimal places earlier by _compute_prf(). To make sure summary.json is as robust as possible for future regression tracking, let’s compute macro_f1 and std_dev from unrounded raw values, and only apply rounding at the very end for the CLI / Markdown outputs.

scripts/benchmark.py
```diff
@@ -333,6 +333,71 @@ def write_run_log(
...
+    stats = compute_aggregate_stats(rows)
+    path = out_dir / "summary.md"
+
+    # metadata
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    git_commit = _get_git_commit()
+    parsers_used = ", ".join(args.parsers)
+
+    # heatmap with top 3 resumes by FP count
+    top_fps = sorted(rows, key=lambda r: r["fp_count"], reverse=True)[:3]
```
Kevin-TCUS, line 344: Minor note: the issue scoped this as top (Resume, Field) combinations. Since rows includes entries across parsers, a single Resume/Field pair could take multiple spots here. I’m okay leaving the logic as-is for this PR, but let’s make sure the PR description matches the actual output behavior.

Kevin-TCUS (changes requested):
Directionally, this is a strong improvement and it hits most of what we were aiming for in #84. Extracting the aggregate summary logic into a reusable helper cleans up the code nicely, and adding summary.md / summary.json as saved run artifacts is a meaningful step forward for both human inspection and future regression tracking.

I’m requesting changes on two points before merge:

1. Let’s compute the aggregate stats once in main() and pass the shared object into the Markdown / JSON / CLI paths rather than recalculating it three times.
2. Let’s compute macro_f1 and std_dev from unrounded raw values, then round only at final output so summary.json is a stronger machine-readable artifact for future comparisons.

Minor note only on the heatmap: I’m okay leaving the current logic as-is for this PR, but the PR description should match the actual output behavior there.

Once those first two points are adjusted, I’m supportive of merging. Nice work getting this framework in place!

Kevin-TCUS (approved):
Looks fantastic, @jriya186 ! The architecture is much cleaner now with the single compute_aggregate_stats call in main(), and the summary.json is generating perfectly.

(Minor note: r["f1"] is actually still rounded upstream by _compute_prf, but the logic you added here gets us exactly what we need for the regression tracking framework to function for v0.3).

Issue #175, opened 2026-04-22
[API] Implement GET /snapshot route using dashboard composition service


Kevin-TCUS:
### Title: `[API] Implement GET /snapshot route using dashboard composition service`

**🔗 Relationship to prior work**
This is the final atomic sub-issue under `#162` (`[API] Implement GET /snapshot`). 

**🎯 Summary**
Wire the completed composition logic into the actual FastAPI route so the frontend can consume it.

**🎯 What we’re doing**
Exposing the `GET /snapshot` endpoint in the backend router, likely in `api/routes/dashboard.py`.

**⚠️ Implementation Boundaries**
* The route should stay extremely thin.
* The route must **not** absorb data selection, summarization, or composition logic. All heavy lifting must remain in the `dashboard_service`.

**🛠️ What you need to do**
* Define `GET /snapshot` in the FastAPI router.
* Call the dashboard composition layer/service.
* Return the reviewer-facing payload defined by the snapshot contract (`#161`).

**✅ How to tell it’s done**
* [ ] `GET /snapshot` exists in the router.
* [ ] Route calls the dashboard composition layer.
* [ ] HTTP response matches the documented contract.
* [ ] Route handler does not contain heavy composition logic.

**🚫 Non-goals**
* Adding auth/gating (handled in a separate security ticket).
* Implementing advanced pagination at this stage.

**🏷️ Labels:** `api`, `backend`, `atomic`
**🏁 Milestone:** `v0.3 – Trustworthy Volunteer Intake`

Issue #174, opened 2026-04-22
[API] Assemble final reviewer-facing snapshot record shape from composed data layers


Kevin-TCUS:
### Title: `[API] Assemble final reviewer-facing snapshot record shape from composed data layers`

**🔗 Relationship to prior work**
This is an atomic sub-issue under `#162` (`[API] Implement GET /snapshot`). 
It uses the helpers built in sub-issues 1-4.

**🎯 Summary**
Build the final one-record-per-submission snapshot payload combining `submissions`, latest parser row, current ops state, and lightweight error summary.

**🎯 What we’re doing**
We are doing the final data assembly to produce the exact JSON shape required by the frontend dashboard. 

**⚠️ Implementation Boundaries**
* **Contract Invariant:** All outputs must conform to the snapshot contract defined in `#161`. This contract is the source of truth for field names and structure.
* **Important Rule:** This assembly step must be a **pure deterministic composition layer**.
* **No I/O:** It must perform **no direct I/O** (no database or Sheets reads). It simply consumes the already-loaded inputs from the previous helpers.
* **Immutability:** The function must **not** mutate input data structures in place; it should construct and return a new output object.

**🛠️ What you need to do**
* Write the master composition function in `dashboard_service.py`.
* Call the helpers (parser selection, ops defaults, error summary).
* Format the final dictionary/JSON output to perfectly match the contract from `#161`.

**✅ How to tell it’s done**
* [ ] Exactly one reviewer-facing record is returned per `submission_id`.
* [ ] Record includes explicit `submission_id`.
* [ ] Raw, parsed, resolved, ops, and error visibility remain distinct.
* [ ] Output shape matches the documented snapshot contract.
* [ ] Assembly is deterministic, pure, and does not mutate inputs.

**🚫 Non-goals**
* Making API/HTTP calls or database reads within this specific function.

**🏷️ Labels:** `api`, `backend`, `atomic`
**🏁 Milestone:** `v0.3 – Trustworthy Volunteer Intake`

Issue #173, opened 2026-04-22
[API] Compose current ops state with default new status handling for snapshot


Kevin-TCUS:
### Title: `[API] Compose current ops state with default new status handling for snapshot`

**🔗 Relationship to prior work**
This is an atomic sub-issue under `#162` (`[API] Implement GET /snapshot`). 

**🎯 Summary**
Attach workflow state to each snapshot record and enforce the default missing-state behavior for new submissions.

**🎯 What we’re doing**
We are merging the `ops` tab data into the read model.
**Core rule:** If no `ops` row exists for a submission, the system must safely default the workflow status to `new`.

**⚠️ Implementation Boundaries**
* **Contract Invariant:** All outputs must conform to the snapshot contract defined in `#161`. This contract is the source of truth for field names and structure.
* This ticket is strictly about read-time composition and default handling. Do not implement `POST` writeback logic here.

**🛠️ What you need to do**
* Write a helper that takes the current `ops` row (if it exists) and formats it for the snapshot.
* Implement the fallback logic so that missing `ops` data yields a valid `new` state object.

**✅ How to tell it’s done**
* [ ] Current `ops` state is attached per `submission_id`.
* [ ] Missing `ops` row yields default status `new`.
* [ ] Fields conform to `#161` and the repo-owned `ops` schema.

**🚫 Non-goals**
* Ops writeback logic or state transition enforcement.

**🏷️ Labels:** `api`, `backend`, `atomic`
**🏁 Milestone:** `v0.3 – Trustworthy Volunteer Intake`

Issue #172, opened 2026-04-22
[API] Derive lightweight per-submission error summary for snapshot composition for 162


Kevin-TCUS:
### Title: `[API] Derive lightweight per-submission error summary for snapshot composition`

**🔗 Relationship to prior work**
This is an atomic sub-issue under `#162` (`[API] Implement GET /snapshot`). 

**🎯 Summary**
Turn the append-only `errors` tab into lightweight error visibility suitable for the reviewer dashboard snapshot.

**🎯 What we’re doing**
Reviewers need to know if a submission failed, but they do not need full backend stack traces in the UI. We are extracting a lightweight summary.
Scope is limited to deriving: `has_error`, latest error summary, and latest error stage/code metadata.

**⚠️ Implementation Boundaries**
* **Contract Invariant:** All outputs must conform to the snapshot contract defined in `#161`. This contract is the source of truth for field names and structure.
* **Latest Error Rule:** "Latest error" is determined by max `created_at` per `submission_id`.
* **Strict Guardrail:** Do **not** include full error payloads, stack traces, or raw error history in the snapshot layer.

**🛠️ What you need to do**
* Write a helper that evaluates the error rows for a given `submission_id`.
* Return a lightweight dictionary/object containing only the summary flags needed by the UI.

**✅ How to tell it’s done**
* [ ] Error presence can be derived per `submission_id`.
* [ ] Latest error is determined by max `created_at`.
* [ ] Full error history is not embedded in the snapshot.
* [ ] The no-error case is treated as a normal, valid state.

**🚫 Non-goals**
* Building a full observability or log-inspection system.
* Exposing raw error logs to the frontend.

**🏷️ Labels:** `api`, `backend`, `atomic`
**🏁 Milestone:** `v0.3 – Trustworthy Volunteer Intake`

PR #189, opened 2026-04-23
[API] Load submissions records for reviewer snapshot composition for issue 170

### LINKED ISSUE #170: [API] Load submissions records for reviewer snapshot composition for 162
### Title: `[API] Load submissions records for reviewer snapshot composition`

**🎯 Summary**
Implement the read step that loads canonical intake rows from the `submissions` tab so they can be used for snapshot generation.

**🎯 What we’re doing**
We are fetching the base user-entered data layer. This helper will pull the raw records and prepare them to be merged with parser, ops, and error data in later steps.

* This ticket is about loading canonical intake data **only**. It must not perform parser selection, ops composition, error summarization, or final snapshot assembly.

**🛠️ What you need to do**
* Write a backend helper to load the `submissions` rows required for the snapshot.
* **Data Structure:** Return shape should be a dictionary/map keyed by `submission_id`, not a raw list.
* Ensure missing or blank optional fields are handled safely without throwing exceptions.

**✅ How to tell it’s done**
* [ ] Reviewer snapshot composition can load `submissions` rows.
* [ ] Return shape is a dictionary/map keyed by `submission_id`.
* [ ] Field names conform to `#161` and the repo-owned schema.
* [ ] Missing/blank optional fields are handled safely.

**🚫 Non-goals**
* Data transformation or composition logic.
* API route wiring.

jliu233:
## 🔗 Related Issues

Closes #170
Part of #162 (GET /snapshot implementation)
Epic: #167

---

## 🎯 Summary

Implements the submissions data loading step for the reviewer snapshot pipeline.

This PR introduces a storage-layer helper that reads canonical intake records from the `submissions` tab and returns them in a dictionary keyed by `submission_id`. This enables downstream snapshot composition without exposing raw Sheets behavior.

---

## 🧠 What This Does

* Adds `load_submission_records()` in `backend/storage/sheets_repo.py`
* Reads all rows from the `submissions` tab
* Maps rows to schema-aligned dictionaries using `sheet_schema`
* Returns:

  ```python
  Dict[str, Dict[str, str]]  # keyed by submission_id
  ```
* Normalizes missing / blank values to `""` to ensure safe downstream usage
* Skips rows without a valid `submission_id`

---

## 📐 Contract Alignment

* Field names strictly follow the repo-owned `submissions` schema (`sheet_schema.py`)
* Output shape matches the requirement for snapshot composition input (dict keyed by `submission_id`)
* No transformation beyond safe normalization is performed

---

## ⚠️ Scope Boundaries (Enforced)

This PR intentionally does **only one thing**: load canonical submissions data.

It does **NOT**:

* select parser results (#171)
* compose ops state (#173)
* summarize errors (#172)
* assemble snapshot records (#174)
* implement API routes (#175)

---

## 🛠️ Implementation Notes

* Uses `get_headers("submissions")` to stay schema-driven
* Pads rows to match schema length to avoid index errors
* Converts all values to strings and trims whitespace
* Deterministic behavior: last occurrence wins if duplicate `submission_id` appears

---

## ✅ How to Test

* Run against a sheet with seeded submissions data
* Call:

  ```python
  load_submission_records()
  ```
* Verify:

  * Returns a dictionary (not a list)
  * Keys are valid `submission_id`s
  * Values contain all submission fields
  * Missing optional fields are `""`, not `None`
  * No exceptions are raised for incomplete rows

---

## 📦 Example Output

```python
{
  "sub_123": {
    "submission_id": "sub_123",
    "created_at": "04-20-2026 12:00:00 UTC",
    "full_name": "Jane Doe",
    "email": "jane@example.com",
    "location_raw": "",
    ...
  }
}
```

---

## 💡 Why This Matters

This is the first step in building the deterministic dashboard read model. It establishes a clean, schema-aligned, backend-owned data source for submissions, which the snapshot composition layer will build on.

### DIFF
--- a/backend/storage/sheets_repo.py
+++ b/backend/storage/sheets_repo.py
@@ -97,7 +97,7 @@ def load_submission_records() -> Dict[str, Dict[str, str]]:
 
     response = _get_sheet().values().get(
         spreadsheetId=GOOGLE_SHEET_ID,
-        range=f"{SUBMISSIONS_SHEET_NAME}!A2:{chr(64 + len(headers))}",
+        range=f"{SUBMISSIONS_SHEET_NAME}!A2:ZZ",
     ).execute()
 
     rows = response.get("values", [])
--- a/backend/storage/sheets_repo.py
+++ b/backend/storage/sheets_repo.py
@@ -6,7 +6,7 @@ from datetime import datetime, timezone
 from googleapiclient.discovery import build
 
 from config import GOOGLE_SHEET_ID
-from sheet_schema import SHEET_SCHEMA, build_row
+from sheet_schema import SHEET_SCHEMA, build_row, get_headers
 from storage._auth import load_service_account_creds, SHEETS_SCOPES
 
 if not GOOGLE_SHEET_ID:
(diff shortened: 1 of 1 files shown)


backend/storage/sheets_repo.py
```diff
@@ -74,6 +74,51 @@ def fetch_live_schema() -> Dict[str, Any]:
...
+        - Does NOT assemble final snapshot records
+
+    Missing/blank cells are normalized to "" so optional fields are handled safely.
+    Rows with no submission_id are skipped because submission_id is the required key.
+    """
+    headers = get_headers(SUBMISSIONS_SHEET_NAME)
+
+    response = _get_sheet().values().get(
+        spreadsheetId=GOOGLE_SHEET_ID,
+        range=f"{SUBMISSIONS_SHEET_NAME}!A2:{chr(64 + len(headers))}",
```
Kevin-TCUS: This is a clever use of ASCII math, but there’s a hidden spreadsheet trap here that catches a lot of people the first time around.

chr(64 + len(headers)) works up through 26 columns (Z). But once the schema grows past that, spreadsheet column naming rolls into AA, AB, etc., while ASCII just keeps going to [ and other characters. At that point the Sheets API range becomes invalid.

The good news is we don’t need a full column-letter converter here. Since your zip/padding logic below already bounds the data safely to len(headers), we can just over-request the range. Something like:

range=f"{SUBMISSIONS_SHEET_NAME}!A2:ZZ"

would make this future-proof without adding extra complexity.

Kevin-TCUS (changes requested):
This is an exemplary PR, @jliu233 . The description is incredibly clear, and you held the architectural boundaries really well. This does exactly what it needs to do for the dashboard composition layer without leaking scope. I also really like the row padding logic, that’s a classic Google Sheets API gotcha that you handled proactively.

I caught one small hidden trap in the range calculation that we should future-proof before we merge. It’s a very common “first time working with spreadsheet APIs” kind of issue, so I left an inline note explaining the why behind it.

Tighten that up and I’ll get this approved. Strong work here.

Kevin-TCUS (approved):
Perfect fix, Jiaqi! The !A2:ZZ range combined with your strict schema-zipping logic makes this read step bulletproof.

The architectural boundaries here are spot on. Approving and merging now so we can keep the dashboard epic moving!

Issue #167, opened 2026-04-21
[Epic] Reviewer Dashboard MVP over the v0.3 4-tab schema


Kevin-TCUS:
### Title: `[Epic] Reviewer Dashboard MVP over the v0.3 4-tab schema`

**🔗 Relationship to prior work**
This epic builds directly on the v0.3 pipeline, schema, and backend hardening work already in place, especially:
* `Refactor backend toward the v0.3 modular architecture` #74
* `Add startup schema validation using backend/sheet_schema.py` #98
* `Extract services/ layer: intake and parser orchestration` #146
* `Extract api/routes/ — thin main.py to startup-only` #156
* `Add repo-owned ops status constants and validation helper` #135
* Any UUID / append-only / deterministic naming tickets that complete the v0.3 pipeline identity model

With the backend pipeline now writing into the 4-tab v0.3 Sheets model, the next missing layer is the actual internal product surface. This epic creates the minimum reviewer dashboard needed to move Libelle from “pipeline + Sheets” into a usable internal tool.

**🎯 What we’re doing**
We are building a lightweight internal dashboard over the existing v0.3 Sheets-backed workflow.
This dashboard is not a new system of record. It is a reviewer-facing decision surface built on top of the current 4-tab schema:
* `submissions`
* `parser_results`
* `ops`
* `errors`

The purpose of this work is to let a reviewer inspect incoming submissions, understand raw user-entered data, inspect parsed and resolved outputs without losing traceability, update workflow state, and work from Libelle instead of directly from raw Google Sheets.

**🏗️ The Target Architecture**
* **`api/routes/dashboard.py`**: Exposes dashboard-facing read/write endpoints such as `GET /snapshot` and the minimum `ops` write endpoint.
* **`services/dashboard_service.py`**: Acts as the dashboard composition layer. Reads the Sheets-backed tabs through `storage/`, applies the v0.3 composition rules, and returns one reviewer-facing record per `submission_id`.
* **`frontend/`**: A lightweight internal dashboard UI. For v0.3, this is a reviewer workflow surface, not a full CRM, matching engine, or analytics platform.

**🧠 The v0.3 Dashboard Model**
The dashboard must reflect the real nature of the underlying data sources. These sources are not equal and should not be treated as a symmetric merge. They are asymmetric:
* `submissions` = canonical user-entered intake record
* `parser_results` = append-only derived interpretation history
* `ops` = mutable workflow state
* `errors` = append-only diagnostic event log

[...]

Issue #162, opened 2026-04-21
API] Implement GET /snapshot over submissions, parser_results, ops, and lightweight error visibility


Kevin-TCUS:
### Title: `[API] Implement GET /snapshot over submissions, parser_results, ops, and lightweight error visibility`

**🔗 Relationship to prior work**
This ticket is the backend implementation step that follows the reviewer-facing `GET /snapshot` contract definition under the `[Epic] Reviewer Dashboard MVP over the v0.3 4-tab schema`.
It builds directly on:
* `Libelle v0.3 Milestone Brief, Architecture, and Execution Map` #76
* `Refactor backend toward the v0.3 modular architecture` #74
* `Add startup schema validation using backend/sheet_schema.py` #98
* `Add repo-owned ops status constants and validation helper` #135
* `[Data] Seed synthetic v0.3 dashboard data across submissions, parser_results, ops, and errors`
* `[API] Define the reviewer-facing GET /snapshot contract for the v0.3 dashboard`

Once the contract is defined, the backend needs to expose a real reviewer-facing read model so the dashboard can consume one stable API instead of reconstructing state from raw Sheets data.

**🎯 Summary**
Implement the `GET /snapshot` endpoint that composes the v0.3 4-tab Sheets-backed workflow into one reviewer-facing record per `submission_id`.
This ticket is not about building a new system of record. It is about implementing the deterministic composition layer described in the snapshot contract so the dashboard has a stable read surface.

**🎯 What we’re doing**
We are implementing a backend read model over the existing v0.3 schema:
* `submissions`
* `parser_results`
* `ops`
* `errors`

The endpoint must:
* return exactly one record per `submission_id`
* apply the parser row selection rule consistently
* tolerate partial pipeline state
* preserve raw vs parsed vs resolved vs workflow vs error distinctions
* provide the dashboard with a clean reviewer-facing view without exposing raw Google Sheets behavior to the frontend

**🏗️ Expected Architecture Direction**
The implementation should live in the dashboard layer, not inside the frontend and not as ad hoc logic in routes.
* **`api/routes/dashboard.py`**: exposes `GET /snapshot`
* **`services/dashboard_service.py`**: performs the composition logic
* **`storage/`**: remains responsible for reading the underlying tabs

*The frontend must not perform composition logic itself.*

**📐 Composition Rules**
The implementation must follow these rules strictly:
* **One record per submission:** Return exactly one reviewer-facing record per `submission_id`.
[...]

PR #190, opened 2026-04-24
add helper function and unit test for issue 171

### LINKED ISSUE #171: [API] Select the latest parser_results row per submission_id for snapshot composition for 162
### Title: `[API] Select the latest parser_results row per submission_id for snapshot composition`

**🎯 Summary**
Implement the deterministic rule for selecting the correct parser row when multiple runs exist for a single submission.

**🎯 What we’re doing**
Because `parser_results` is append-only, we must programmatically select the active row for the dashboard. 
**Core rule:** Choose the row with the highest `parser_run_id`. Break ties by latest `created_at`.

* **Empty-state rule:** If no `parser_results` rows exist for a `submission_id`, the function must return a null/empty-state result. This is a valid state and must **not** raise an error or block assembly.

**🛠️ What you need to do**
* Write a pure, deterministic helper function that takes a list of parser rows and returns the latest one according to the rules above.
* Write a unit test proving it selects the highest run ID and handles the empty state correctly.

**✅ How to tell it’s done**
* [ ] Given multiple `parser_results` rows for one `submission_id`, the helper returns the correct latest row.
* [ ] Behavior is deterministic and covered by tests.
* [ ] Missing parser rows are handled as a valid state.
* [ ] No snapshot-fatal behavior is introduced for parser-pending submissions.

**🚫 Non-goals**
* Final snapshot assembly.
* Modifying parser behavior or schema.

jliu233:
## 🔗 Related Issues

Closes #171
Part of #162 (GET /snapshot implementation)
Epic: #167

---

## 🎯 Summary

Implements the deterministic selection rule for identifying the active `parser_results` row per `submission_id`.

This PR introduces a pure helper that selects the correct parser run from an append-only history, enabling consistent downstream snapshot composition.

---

## 🧠 What This Does

* Adds `select_latest_parser_result()` in `backend/services/dashboard_parser_results.py`
* Applies the selection rule:

  1. Choose the row with the highest `parser_run_id`
  2. Break ties using the latest `created_at`
* Returns:

  ```python
  Optional[Dict[str, Any]]
  ```
* Handles empty state safely by returning `None` (valid parser-pending state)
* Includes unit tests to validate deterministic behavior

---

## 📐 Contract & Behavior Guarantees

* Deterministic: same input → same output
* No mutation of input data
* Compatible with append-only `parser_results` design
* Supports partial pipeline state (no parser rows is valid)
* Does not raise exceptions for missing data

---

## ⚠️ Scope Boundaries (Enforced)

This PR is intentionally narrow and only handles parser row selection.

It does **NOT**:

* load parser_results from Sheets
* assemble snapshot records (#174)
* apply ops defaults (#173)
* summarize errors (#172)
* implement API routes (#175)

---

## 🛠️ Implementation Notes

* Numeric `parser_run_id` values are compared numerically (e.g., `"10" > "2"`)
* Non-numeric IDs fall back to string comparison
* `created_at` supports current repo timestamp format and common ISO formats
* Falls back to string comparison if parsing fails
* Returns a copy of the selected row to avoid accidental mutation

---

## ✅ How to Test

Run:

```bash
pytest backend/tests/test_dashboard_parser_results.py
```

Covers:

* selecting highest `parser_run_id`
* tie-breaking with latest `created_at`
* empty input returns `None`
* deterministic behavior across runs

---

## 📦 Example

```python
rows = [
  {"parser_run_id": "1", "created_at": "04-20-2026 10:00:00 UTC"},
  {"parser_run_id": "10", "created_at": "04-20-2026 09:00:00 UTC"},
]

select_latest_parser_result(rows)
# → returns row with parser_run_id = "10"
```

### DIFF
--- a/backend/services/dashboard_parser_results.py
+++ b/backend/services/dashboard_parser_results.py
@@ -1,0 +1,77 @@
+"""Pure parser_results selection helpers for dashboard snapshot composition."""
+
+from datetime import datetime
+from typing import Any, Dict, List, Optional, Tuple
+
+
+ParserResultRow = Dict[str, Any]
+
+
+def _parser_run_sort_value(parser_run_id: Any) -> Tuple[int, Any]:
+    """
+    Build a deterministic sort value for parser_run_id.
+
+    Numeric-looking IDs are sorted numerically so "10" is newer than "2".
+    Other IDs are sorted as strings.
+    """
+    value = "" if parser_run_id is None else str(parser_run_id).strip()
+
+    if value.isdigit():
+        return (1, int(value))
+
+    return (0, value)
+
+
+def _created_at_sort_value(created_at: Any) -> Tuple[int, Any]:
+    """
+    Build a deterministic sort value for created_at.
+
+    Supports the current repo timestamp format and common ISO-like timestamps.
+    Falls back to string sorting if parsing fails.
+    """
+    value = "" if created_at is None else str(created_at).strip()
+
+    for fmt in (
+        "%m-%d-%Y %H:%M:%S %Z",
+        "%Y-%m-%dT%H:%M:%S%z",
+        "%Y-%m-%dT%H:%M:%S",
+        "%Y-%m-%d %H:%M:%S",
+    ):
+        try:
+            return (1, datetime.strptime(value, fmt))
+        except ValueError:
+            continue
+
+    return (0, value)
+
+
+def select_latest_parser_result(
+    parser_rows: List[ParserResultRow],
+) -> Optional[ParserResultRow]:
+    """
+    Select the latest parser_results row for one submission_id.
+
+    Selection rule:
+        1. Choose the row with the highest parser_run_id.
+        2. If parser_run_id ties, choose the latest created_at.
+
+    Empty state:
+        If no parser_results rows exist, return None. This is a valid
+        parser-pending state and should not block snapshot assembly.
+
+    Scope boundary:
+        This helper does not read from Sheets, assemble snapshots, modify parser
+        behavior, compose ops state, or summarize errors.
+    """
+    if not parser_rows:
+        return None
+
+    latest_row = max(
+        parser_rows,
+        key=lambda row: (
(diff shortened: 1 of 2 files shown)


Kevin-TCUS (changes requested):
Jiaqi, this is really strong work. You nailed the contract for #171 and kept the helper scoped exactly where it belongs.

A few things that stood out:

* catching the string-sort issue for parser_run_id and comparing numeric-looking IDs numerically. Sorting "10" vs "2" is a classic bug, and you handled it proactively.
* returning dict(latest_row) so downstream snapshot code does not accidentally mutate the original row.
* keeping this purely focused on parser row selection, with no Sheets loading, snapshot assembly, ops logic, or route work.

I’m requesting one small change before merge because of a subtle Python timestamp comparison issue.

_created_at_sort_value() currently returns parsed datetime objects. One supported format (%z) can produce a timezone-aware datetime, while the other formats produce timezone-naive datetimes. If two rows tie on parser_run_id, and Python tries to compare a timezone-aware datetime against a timezone-naive datetime, it can raise:

TypeError: can't compare offset-naive and offset-aware datetimes

Let’s normalize parsed timestamps into one comparable representation, such as numeric UTC timestamps. For example:

```python
dt = datetime.strptime(value, fmt)
if dt.tzinfo is None:
    dt = dt.replace(tzinfo=timezone.utc)
return (1, dt.timestamp())
```

You’ll need to import timezone from datetime.

I’d also like one small test for the defensive-copy behavior, since the PR explicitly guarantees that the selected row is copied before return:

```python
def test_select_latest_parser_result_returns_copy() -> None:
    row = {
        "submission_id": "sub_001",
        "parser_run_id": "1",
        "created_at": "04-20-2026 10:00:00 UTC",
    }

    latest = select_latest_parser_result([row])

    assert latest is not row
    latest["parser_run_id"] = "999"
    assert row["parser_run_id"] == "1"
```

Once those two small updates are in, I’m supportive of merging. The structure and scope discipline here are very good.

Kevin-TCUS (approved):
Nice work!!

Issue #161, opened 2026-04-21
[API] Define the reviewer-facing GET /snapshot contract for the v0.3 dashboard


Kevin-TCUS:
### Title: `[API] Define the reviewer-facing GET /snapshot contract for the v0.3 dashboard`

**🔗 Relationship to prior work**
This ticket is the contract-definition step under the `[Epic] Reviewer Dashboard MVP over the v0.3 4-tab schema`.
It builds directly on:
* `Libelle v0.3 Milestone Brief, Architecture, and Execution Map` #76
* `Refactor backend toward the v0.3 modular architecture` #74
* `Add startup schema validation using backend/sheet_schema.py` #98
* `Add repo-owned ops status constants and validation helper` #135
* `[Data] Seed synthetic v0.3 dashboard data across submissions, parser_results, ops, and errors` (The seed-data ticket ensuring realistic synthetic records exist)

Before frontend work and backend implementation proceed in parallel, the project needs a stable reviewer-facing snapshot contract. Without that contract, backend and frontend will improvise different assumptions and the dashboard lane will drift.

**🎯 Summary**
Define the JSON contract returned by `GET /snapshot` so the v0.3 dashboard can render a reviewer-facing view over the 4-tab Sheets-backed schema.
This ticket is not the endpoint implementation itself. It is the contract-definition step that locks the structure, layering, and composition rules before the backend route and frontend UI are built against it.

**🎯 What we’re doing**
We are defining a reviewer-facing read model over the current v0.3 data system.
That read model is not a new source of truth and it is not a symmetric merge of equal tabs.
It is a deterministic composition layer over four asymmetric system-of-record sources:
* `submissions` = canonical user-entered intake record
* `parser_results` = append-only derived interpretation history
* `ops` = mutable workflow state
* `errors` = append-only diagnostic event log

The purpose of this contract is to give the dashboard one stable interface that:
* preserves trust in the source data
* tolerates partial pipeline state
* keeps workflow state usable
* avoids leaking raw Sheets logic into the frontend

**📋 What the contract must define**
The contract must define a concrete JSON response shape for `GET /snapshot`.
At minimum, it must define:
* top-level snapshot metadata
* exactly one reviewer-facing record per `submission_id`
* explicit grouping of data by layer, such as:
  * submission fields
  * parser fields
  * resolver / derived fields
  * ops fields
  * lightweight error summary

The contract must also explicitly identify or group which fields are intended for:
* Inbox / list view
* detail view

[...]

PR #192, opened 2026-04-28
add new helper for lightweight per-submission error summary

### LINKED ISSUE #172: [API] Derive lightweight per-submission error summary for snapshot composition for 162

jliu233:
## 🔗 Related Issue

Closes #172
Part of #162 (GET `/snapshot`)

---

## 🎯 Summary

Introduces a **pure helper function** to derive lightweight, per-submission error visibility from the append-only `errors` tab.

This enables the reviewer dashboard to determine:

* whether a submission has failed
* what the latest error is (summary only)
* where it failed (stage/code)

without exposing raw logs or full error history.

---

### New helper module

**`backend/services/dashboard_errors.py`**

Adds:

```python
summarize_submission_errors(submission_id, error_rows)
```

This function:

* Filters error rows by `submission_id`
* Selects the **latest error using `max(created_at)`**
* Returns a **lightweight summary object**:

```python
{
  "has_error": bool,
  "latest_error_summary": str,
  "latest_error_stage": str,
  "latest_error_code": str,
}
```

---

### Timestamp handling (consistency with #171)

* Reuses the same `_created_at_sort_value` pattern as `dashboard_parser_results.py`
* Normalizes:

  * timezone-aware timestamps
  * timezone-naive timestamps
* Prevents Python comparison issues (`offset-naive vs offset-aware`)

---

### Unit tests

**`backend/tests/test_dashboard_errors.py`**

Covers:

* ✅ No-error case (empty + no matching submission)
* ✅ Correct latest error selection by `created_at`
* ✅ Proper filtering by `submission_id`
* ✅ Guardrail: no exposure of `error_details` or history
* ✅ Timestamp normalization (naive vs aware)

---

## ✅ Contract Compliance

* **Latest Error Rule**: enforced via `max(created_at)`
* **Snapshot Contract (#161)**: only returns required lightweight fields
* **Strict Guardrail**:

  * ❌ No stack traces
  * ❌ No `error_details`
  * ❌ No raw history

---

## 🧪 How to test

```bash
pytest backend/tests/test_dashboard_errors.py
```

---

## 🚫 Non-goals (intentionally excluded)

* No Sheets access or loading logic
* No snapshot assembly
* No observability/log inspection features
* No mutation of error data

### DIFF
--- a/backend/services/dashboard_errors.py
+++ b/backend/services/dashboard_errors.py
@@ -1,0 +1,76 @@
+"""Pure error summary helpers for dashboard snapshot composition."""
+
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional, Tuple
+
+
+ErrorRow = Dict[str, Any]
+ErrorSummary = Dict[str, Any]
+
+
+def _created_at_sort_value(created_at: Any) -> Tuple[int, Any]:
+    """
+    Build a deterministic sort value for created_at.
+
+    Supports the current repo timestamp format and common ISO-like timestamps.
+    Falls back to string sorting if parsing fails.
+    """
+    value = "" if created_at is None else str(created_at).strip()
+
+    for fmt in (
+        "%m-%d-%Y %H:%M:%S %Z",
+        "%Y-%m-%dT%H:%M:%S%z",
+        "%Y-%m-%dT%H:%M:%S",
+        "%Y-%m-%d %H:%M:%S",
+    ):
+        try:
+            dt = datetime.strptime(value, fmt)
+            if dt.tzinfo is None:
+                dt = dt.replace(tzinfo=timezone.utc)
+            return (1, dt.timestamp())
+        except ValueError:
+            continue
+
+    return (0, value)
+
+
+def summarize_submission_errors(
+    submission_id: str,
+    error_rows: List[ErrorRow],
+) -> ErrorSummary:
+    """
+    Derive lightweight error visibility for one submission_id.
+
+    Selection rule:
+        Latest error is the row with the highest created_at for the submission_id.
+
+    Empty state:
+        If no matching error rows exist, return a normal no-error summary.
+
+    Scope boundary:
+        This helper does not read from Sheets, assemble snapshots, expose raw
+        error details, include stack traces, or return full error history.
+    """
+    matching_rows = [
+        row for row in error_rows if str(row.get("submission_id", "")) == submission_id
+    ]
+
+    if not matching_rows:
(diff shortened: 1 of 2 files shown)


Kevin-TCUS (approved):
Approved! This is really strong work, @jliu233 . 

This PR matches the #172 contract perfectly: it derives lightweight error visibility, safely selects the latest error, handles the no-error state gracefully, and strictly protects against exposing raw stack traces.

A few things I especially appreciate about this PR:
1. **Applied Learning:** You carried forward the timezone-aware vs. timezone-naive timestamp lesson from the previous parser PR and normalized them safely here *before* I even asked. That is fantastic engineering growth.
2. **Strict Boundaries:** The helper stays completely pure—no Sheets I/O, no route logic, just data in and data out.
3. **Comprehensive Testing:** Your tests explicitly prove the guardrails work (especially `test_summarize_submission_errors_does_not_expose_error_details_or_history`). 

**One tiny, non-blocking tip for the future:**
In `matching_rows`, you safely cast the row's ID: `str(row.get("submission_id", "")) == submission_id`. Since Python doesn't strictly enforce type hints at runtime, a super defensive habit is to cast both sides just in case a higher layer accidentally passes an int: `... == str(submission_id)`. Absolutely no need to change it here since the contract expects a string, just a handy habit to keep in your back pocket!

Excellent job. Merging now!

PR #195, opened 2026-05-01
Compose current ops state with default new status handling for snapshot (issue 173)

### LINKED ISSUE #173: [API] Compose current ops state with default new status handling for snapshot

jliu233:
## Summary

Closes #173.

Adds pure backend helpers for composing the current ops workflow state into snapshot records. If a submission has no matching ops row, the snapshot read model receives a valid default ops state with `status: "new"` and blank repo-owned ops fields.

## Changes

* Added `format_current_ops_state()` to normalize an ops row for snapshot output.
* Added `compose_current_ops_state()` to select the current ops row by `submission_id`.
* Defaults missing ops data to the contract-safe `"new"` workflow state.
* Guards against invalid ops statuses by falling back to `"new"`.
* Keeps output limited to repo-owned ops schema fields and excludes `submission_id` from the nested state object.
* Added focused tests for:
  * missing ops row fallback
  * schema-shaped formatting
  * invalid status fallback
  * filtering by `submission_id`
  * selecting the latest matching ops row

## Non-Goals

* No ops writeback logic.
* No workflow transition enforcement.
* No mutation of ops rows or sheet data.

### DIFF
--- a/backend/services/dashboard_ops_state.py
+++ b/backend/services/dashboard_ops_state.py
@@ -1,0 +1,102 @@
+"""Pure ops state helpers for dashboard snapshot composition."""
+
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional, Tuple
+
+from ops_schema import OPS_STATUS_NEW, is_valid_ops_status
+from sheet_schema import OPS_HEADERS
+
+
+OpsRow = Dict[str, Any]
+OpsState = Dict[str, Any]
+
+OPS_STATE_FIELDS = tuple(field for field in OPS_HEADERS if field != "submission_id")
+
+
+def _updated_at_sort_value(updated_at: Any) -> Tuple[int, Any]:
+    """
+    Build a deterministic sort value for updated_at.
+
+    Supports the current repo timestamp format and common ISO-like timestamps.
+    Falls back to string sorting if parsing fails.
+    """
+    value = "" if updated_at is None else str(updated_at).strip()
+
+    for fmt in (
+        "%m-%d-%Y %H:%M:%S %Z",
+        "%Y-%m-%dT%H:%M:%S%z",
+        "%Y-%m-%dT%H:%M:%S",
+        "%Y-%m-%d %H:%M:%S",
+    ):
+        try:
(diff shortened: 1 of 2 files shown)


Kevin-TCUS (approved):
@jliu233 , this looks very aligned with #173. Nice work keeping this as a pure read-model helper rather than mixing in writeback, transition enforcement, route logic, or Sheets access.

The implementation matches the snapshot boundary perfectly:
* missing ops row returns a complete default status: `"new"` state
* invalid statuses fall back safely to `"new"`
* output is limited to repo-owned ops fields
* `submission_id` and unowned fields are not exposed in the nested ops state
* current state selection is deterministic using latest `updated_at`
* tests cover the important cases directly

I also really like that `_updated_at_sort_value()` follows the same defensive timestamp pattern we’ve been using elsewhere: parse known timestamp formats first, fall back to deterministic string sorting if needed. Great job carrying that pattern forward.

**One tiny non-blocking note:** `compose_current_ops_state()` strips the row `submission_id`, but not the input `submission_id`. Since our IDs should be system-generated and clean, I do not think that matters here, but it's a good edge case to keep in mind!

This is the exact right scope for #173. Approving and merging!

PR #197, opened 2026-05-04
Implements the pure snapshot assembly layer for issue 174

### LINKED ISSUE #174: [API] Assemble final reviewer-facing snapshot record shape from composed data layers

jliu233:
## Summary

This PR adds `services/dashboard_service.py` with a master composition function that combines already-loaded submissions, parser results, ops rows, and error rows into one reviewer-facing snapshot record per `submission_id`.

## What Changed

- Added `assemble_snapshot_records(...)`
- Preserves distinct snapshot layers:
  - `raw`
  - `parsed`
  - `resolved`
  - `ops`
  - `errors`
- Calls existing helper layers:
  - latest parser result selection
  - current ops state/defaults
  - lightweight error summary
- Adds resolver state visibility:
  - `not_run`
  - `resolved`
  - `zero_matches`
- Keeps the assembly function pure:
  - no Sheets/database/API I/O
  - deterministic ordering
  - no input mutation

## Tests

Added `backend/tests/test_dashboard_service.py` covering:

- one record per `submission_id`
- expected layered record shape
- latest parser row usage
- missing parser/ops/error defaults
- resolver not-run vs zero-match distinction
- input immutability

## Verification

- `py_compile` passed for the new service and tests
- Direct invocation of the new test functions passed
- Full `pytest` was not run because `pytest` is not installed in the local venv

Closes #174

### DIFF
--- a/backend/services/dashboard_service.py
+++ b/backend/services/dashboard_service.py
@@ -1,0 +1,165 @@
+"""Pure dashboard snapshot composition service."""
+
+import json
+from typing import Any, Dict, List, Mapping, Optional
+
+from services.dashboard_errors import summarize_submission_errors
+from services.dashboard_ops_state import compose_current_ops_state
+from services.dashboard_parser_results import select_latest_parser_result
+from sheet_schema import SUBMISSIONS_HEADERS
+
+
+SubmissionRecord = Mapping[str, Any]
+ParserResultRow = Mapping[str, Any]
+OpsRow = Mapping[str, Any]
+ErrorRow = Mapping[str, Any]
+SnapshotRecord = Dict[str, Any]
+
+RAW_FIELDS = tuple(field for field in SUBMISSIONS_HEADERS if field != "submission_id")
+PARSED_FIELDS = (
+    "parser_run_id",
+    "created_at",
+    "parser_version",
+    "parsed_skills_raw",
+    "parsed_location_raw",
+    "parser_confidence",
+)
+RESOLVED_FIELDS = (
+    "resolver_version",
+    "aliases_version",
+    "resolved_skill_ids",
+    "unknown_skills",
+    "resolver_coverage",
+)
+
+
+def assemble_snapshot_records(
+    submissions_by_id: Mapping[str, SubmissionRecord],
+    parser_rows: List[ParserResultRow],
+    ops_rows: List[OpsRow],
+    error_rows: List[ErrorRow],
+) -> List[SnapshotRecord]:
+    """
+    Compose reviewer-facing dashboard snapshot records from preloaded data layers.
+
+    This is a deterministic, side-effect-free composition layer. It performs no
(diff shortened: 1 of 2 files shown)


Kevin-TCUS (approved):
@jliu233 , this is very strong and lines up perfectly with #174.

This PR is exactly the snapshot assembly layer we needed: it takes the already-loaded submissions, parser rows, ops rows, and error rows, and cleanly produces one reviewer-facing record per `submission_id` while keeping the layers distinct.

I especially appreciate that the function stays pure and deterministic:
* No Sheets / database / API I/O
* No input mutation (great job writing a test to explicitly prove this!)
* Raw / parsed / resolved / ops / errors remain safely separated
* Missing parser / ops / error states are handled gracefully as normal snapshot states
* Resolver `not_run` vs `zero_matches` is explicitly and safely represented

The tests are also extremely well targeted. The `not_run` vs `zero_matches` case is specifically important for frontend clarity because those are meaningfully different states for our reviewers.

**One small, non-blocking performance tip for the future:** It looks like `ops_rows` and `error_rows` are scanned inside the loop for each submission. That is 100% fine for our v0.3 scale! However, as a general pattern when datasets grow large, it's often safer to pre-group those lists into dictionaries (keyed by `submission_id`) *before* the loop. That way, you get instant O(1) lookups inside the loop instead of O(N^2) list scans. Absolutely no need to change it here, just a handy pattern to keep in your back pocket!

This is the exact right scope and boundary for #174. Approving and merging! We are officially ready to wire up the route.

Issue #198, opened 2026-05-04
[Benchmark/API] Add shared block-based PDF text extraction helper for intake and benchmark


Kevin-TCUS:
# [Benchmark/API] Add shared block-based PDF text extraction helper for intake and benchmark

## Context
Recent parser work under #149 identified that PyMuPDF’s default `get_text("text")` extraction can return PDF text in internal document storage order rather than visual reading order.

This can scramble resumes before the parser ever sees them, especially on layout-sensitive PDFs. When that happens, parser benchmark scores may look bad even though the deeper issue is the PDF-to-text extraction step.

Riya tested a block-based PyMuPDF extraction approach using `get_text("blocks")` plus coordinate/spatial sorting. That approach appears to improve text order for affected resumes by sorting extracted text blocks top-to-bottom and left-to-right before passing text into the parser.

**Current issue:**
Both the production intake path and `scripts/benchmark.py` should use the same PDF text extraction behavior.

### In plain language:
```text
PDF
→ shared PDF text extraction helper
→ clean extracted text
→ parse_resume(...)  
```
**Used by:**
* `intake_service.py`
* `scripts/benchmark.py`

If production and benchmark extract text differently, benchmark scores may not reflect the same parser input that production uses.

## Goal
Implement one shared PDF text extraction helper and use it from both:
* the production intake/parser path
* `scripts/benchmark.py`

The helper should use the improved block-based PyMuPDF extraction approach identified in #149.
*This issue is about aligning the PDF-to-text step. It is not about changing parser scoring or parser logic.*

## What a shared helper means
A helper is one small reusable function that does one job. In this case, the helper’s job is:
*“Given a PDF, extract text using Libelle’s approved PDF extraction method.”*

Instead of having PDF extraction logic copied separately in:
* `intake_service.py`
* `scripts/benchmark.py`

we want one shared helper, for example: `extract_text_from_pdf(...)`.

Then both production intake and benchmark call the same extraction logic. This keeps production and benchmark aligned. If we improve PDF extraction later, we only update one place.

## Suggested implementation shape
Exact module location can be chosen during implementation, but one possible shape is:
`backend/services/pdf_text_extraction.py`
with helper functions such as:
`extract_text_from_pdf_bytes(pdf_bytes: bytes) -> str`
and/or
`extract_text_from_pdf_path(pdf_path: Path) -> str`

**Why two helpers may be useful:**
* `intake_service.py` works from uploaded PDF bytes
[...]

PR #199, opened 2026-05-05
[Benchmark/API] Add shared block-based PDF text extraction helper

jriya186:
## Summary

This PR factors PyMuPDF PDF text extraction into a single shared helper module
(`backend/services/pdf_text_extraction.py`) and updates both the production intake
path and `scripts/benchmark.py` to use it. This PR is fixing the issue #198 

Prior to this change, `intake_service.py` and `benchmark.py` each owned separate PDF
extraction implementations - `intake_service.py` had already been updated to use
block-based extraction, but `benchmark.py` was still using the old `get_text("text")`
path, meaning production and benchmark were feeding different extracted text into the
same parser.

> **This PR aligns text extraction, not scoring. Benchmark scoring semantics are unchanged.**

---

### New file: `backend/services/pdf_text_extraction.py`
Shared module with three functions:
- `_extract_text_from_fitz_doc(doc)` — internal helper, does block extraction + spatial sort on an open fitz document
- `extract_text_from_pdf_bytes(pdf_bytes)` — used by `intake_service.py` (uploaded PDF bytes)
- `extract_text_from_pdf_path(pdf_path)` — used by `benchmark.py` (local file paths)

### `backend/services/intake_service.py`
- `_extract_text_from_pdf` now delegates to `extract_text_from_pdf_bytes`
- `import fitz` removed — no longer needed
- `IntakeError` wrapping stays in place

### `scripts/benchmark.py`
- `_run_libelle` now calls `extract_text_from_pdf_path` instead of owning its own fitz block

---

## Why Block-Based Extraction

PyMuPDF's default `get_text("text")` returns text in internal PDF storage order, which
can differ from visual reading order. `get_text("blocks")` returns each text block with
its bounding box coordinates. Sorting blocks by `(y0, x0)` before joining restores
top-to-bottom, left-to-right reading order, which is what the parser expects.

---

## Before / After — `low_signal_02`

With the old `get_text("text")` extraction, `low_signal_02` produced scrambled text
where the skills section content appeared **before** its header, resulting in
`skills_lines: []` and an F1 of **0.0**.

With block-based extraction, text follows visual order and the skills section is
correctly identified:

```python
# Before
skills_lines: []  # header never found — content appeared before it in storage order

# After
skills_lines: [
    'Basic Health Education Support',
    'Data Entry and Record Maintenance',
    'Survey Assistance',
    'Written Communication'
]
```
## Before / After — `low_signal_01` (partial recovery)

Block-based extraction unscrambled the text enough for the parser to find all 4 expected
skills (recall: 1.0), but also began capturing contact info, summary text, and education
content as false positives (precision: 0.286, F1: 0.444):

This is a parser-layer section boundary issue - the skills section start is triggering
too early, before the actual `SKILLS` header. Targeted fix tracked in #149.

| Resume | Before | After (this PR) |
|--------|--------|-----------------|
| `low_signal_02` | 0.0 | 1.0 |
| `low_signal_01` | 0.0 | 0.444 (full recovery in #149) |
---

## Acceptance Criteria

- [x] Production intake and benchmark harness use the same PDF text extraction helper
- [x] `intake_service.py` no longer owns a separate PDF extraction implementation
- [x] `scripts/benchmark.py` no longer owns a divergent PDF extraction implementation
- [x] Benchmark scoring semantics unchanged
- [x] Parser logic unchanged
- [x] Before/after notes included for low signal resumes
- [x] No unrelated parser, resolver, storage, or API changes included

---

## Non-goals

Multi-column PDF extraction, parser logic changes, scoring changes, and golden JSON
changes are all out of scope for this PR and exist in their respective issues.

### DIFF
--- a/backend/services/intake_service.py
+++ b/backend/services/intake_service.py
@@ -5,8 +5,6 @@ import traceback
 import uuid
 from typing import Any, Dict, List, Optional, Union
 
-import fitz
-
 from config import MAX_PDF_MB
 from storage.drive_repo import upload_pdf
 from storage.sheets_repo import write_base_row
@@ -78,10 +76,8 @@ def _parse_interests(raw: Union[str, List[str], None]) -> str:
 
 def _extract_text_from_pdf(pdf_bytes: bytes) -> str:
     try:
-        doc = fitz.open(stream=pdf_bytes, filetype="pdf")
-        text = "\n".join([p.get_text("text") for p in doc])
-        doc.close()
-        return text
+        from services.pdf_text_extraction import extract_text_from_pdf_bytes
+        return extract_text_from_pdf_bytes(pdf_bytes)
     except Exception:
         traceback.print_exc()
         raise IntakeError("PDF_PARSE_FAILED", "PDF parsing failed", status_code=400)
--- a/backend/services/pdf_text_extraction.py
+++ b/backend/services/pdf_text_extraction.py
@@ -1,0 +1,67 @@
+"""
+Shared PDF text extraction helper for Libelle.
+
+Provides block-based PyMuPDF extraction with spatial sorting so that
+extracted text follows visual reading order (top-to-bottom, left-to-right)
+rather than internal PDF storage order.
+
+Used by:
+  - services/intake_service.py  (bytes path, from uploaded PDF)
+  - scripts/benchmark.py        (file path, from local PDF files)
+"""
+
+from pathlib import Path
+
+import fitz  # PyMuPDF
+
+
+def _extract_text_from_fitz_doc(doc: fitz.Document) -> str:
+    """
+    Internal helper: extract and spatially sort text blocks from an open
+    PyMuPDF document.
+
+    Each page's text blocks are collected with their bounding box coordinates.
+    Blocks are sorted top-to-bottom (y0), then left-to-right (x0) so the
+    resulting string follows visual reading order rather than the internal
+    PDF storage order fitz uses by default.
+
+    Only text blocks (block type 0) are included — image blocks (type 1)
+    are skipped.
+    """
+    blocks = []
+    for page in doc:
+        blocks.extend(page.get_text("blocks"))
+
+    # sort by vertical position first, then horizontal
+    blocks.sort(key=lambda b: (b[1], b[0]))
+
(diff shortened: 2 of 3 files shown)


Kevin-TCUS (approved):
@jriya186 , this is a really strong direction and beautifully aligned with #198. Factoring the extraction into a shared helper is exactly what we needed to keep the API and the benchmark harness perfectly synced. You also did a great job managing resources safely with the `try...finally` block!

I do have one requested change before we merge this, and it has to do with a tricky quirk of how PDF coordinate systems work.

In `_extract_text_from_fitz_doc`, the helper currently collects blocks across all pages into one list and then sorts them globally by `(y0, x0)`. The catch is that PyMuPDF resets `y0` to `0` at the top of every page. That means if we sort globally, the top of Page 1 and the top of Page 2 will get mashed together, scrambling multi-page resumes.

**The fix:** We need to preserve the document's page order and only sort blocks *within* each individual page. 

Conceptually, the loop should look like this:
* iterate page by page
* get that page’s text blocks
* filter to text blocks only (block type 0)
* sort that specific page’s blocks by `(y0, x0)`
* append the sorted page text to a master string/list
* join the pages together in original document order

Everything else here looks phenomenal. The helper separation is clean, the delegation is right, and you safely avoided changing the parser logic. Once we get that page-order loop updated, this is good to merge!

Kevin-TCUS (changes requested):
Whoops, accidentally hit 'Approve' on my last review instead of 'Request changes'! Still waiting on that page-order fix before we merge

jriya186:
Hi @Kevin-TCUS 
I made the change you requested and tested on a few multi-page resumes and it is working as expected. Please let me know it is up to the mark. Thanks again for the direction!

Kevin-TCUS (approved):
@jriya186 , thank you for making the page-order fix. This addresses the main concern I had with the original implementation.

The helper now keeps the benefit of block-based spatial sorting while preserving document page order, which is exactly what we need for multi-page resumes. I also appreciate that this PR stayed tightly scoped: intake and benchmark now share one extraction helper, while parser logic, scoring semantics, resolver behavior, and gold JSONs remain unchanged.

This is a strong architecture improvement because production intake and the benchmark harness now feed the parser through the same extracted-text path, making benchmark results more representative of the real intake flow.

Approved and good to merge.

PR #200, opened 2026-05-05
Implement GET /snapshot route for issue 175

jliu233:
## Summary

Closes #175

Implements `GET /snapshot` by wiring the existing dashboard snapshot composition service into a thin FastAPI route.

## Changes

- Added `GET /snapshot` route in `backend/api/routes/dashboard.py`
- Registered the dashboard router in `backend/main.py`
- Added `get_snapshot_records()` in `dashboard_service` to orchestrate snapshot loading and composition
- Added schema-aligned loaders for:
  - `parser_results`
  - `ops`
  - `errors`
- Added tests covering:
  - the route returning the dashboard service payload
  - the service loading all snapshot data layers before composition

## Notes

- The route handler intentionally stays thin.
- Data loading and snapshot composition remain in the service/storage layers.
- No auth or gating was added, per the issue non-goals.
- Docs were not updated because API spec updates are out of scope for this issue.

## Testing

- Verified changed backend files compile successfully.
- Verified `/snapshot` is registered and returns the service payload through a `TestClient` check.

PR #202, opened 2026-05-06
Add typed response model for GET /snapshot (for issue 201)

### LINKED ISSUE #201: [API] Add typed response model for GET /snapshot
# [API] Add typed response model for GET /snapshot

## Context
Follow-up to PR #200.

`GET /snapshot` is now wired into the dashboard service and returns composed reviewer-dashboard data. The route is intentionally thin, which is good, but the response shape is still implicit.

For v0.3, the frontend reviewer dashboard will depend on this endpoint as a stable contract. We should make that contract explicit with typed Pydantic response models before more UI work builds on top of it.

## Goal
Add an explicit typed response model for the reviewer-facing `GET /snapshot` payload.

This should make the frontend/backend boundary more stable, reduce accidental response-shape drift, and keep the route thin while allowing FastAPI/Pydantic to validate and document the returned structure.

## What we are doing
Define Pydantic models that represent the current snapshot contract, then attach the top-level model to the `GET /snapshot` route using FastAPI’s `response_model`.

The implementation should reflect the current composed snapshot shape. **This issue is about typing and contract stabilization, not changing what data the snapshot service returns.**

## Why this matters
The reviewer dashboard is moving from backend composition into frontend consumption. Without an explicit response model, the API contract exists only as an informal dict shape in service code and tests.

For v0.3, we want:
- A stable backend/frontend contract.
- Clear expectations for reviewer-facing fields.
- Early detection if snapshot composition accidentally changes shape.
- Better generated API documentation from FastAPI.
- Less guesswork for frontend contributors.

## Implementation boundaries
**The implementation should:**
- Add typed Pydantic models for the snapshot response.
- Keep the route handler thin.
- Declare the `response_model` directly in the `@router.get("/snapshot")` decorator.
- Preserve the existing snapshot composition behavior.
- Match the existing reviewer-facing contract from #161 and PR #200.
- Keep optional fields intentional rather than loosely typed.
- Avoid broad `dict`, `Any`, or untyped nested payloads unless there is a specific reason.
- Keep model names clear enough for frontend/backend contributors to understand.

**The implementation should not:**
- Change the snapshot composition logic.
- Add auth/gating.
- Add pagination.
- Change persistence behavior.
- Rename fields unless required to match the documented contract.
- Introduce frontend code.
- Add new reviewer workflow behavior.

## Suggested implementation shape
Exact names are flexible, but the response should map to the list of dictionaries currently returned by the service. The models should represent the nested pieces of each submission:

```python
class SnapshotRawData(BaseModel):
    full_name: str
    skills_raw: str
    # ... other raw fields

class SnapshotParsedData(BaseModel):
    parser_state: str
    parser_run_id: Optional[str]
    # ... other parsed fields

class ReviewerSubmissionSnapshot(BaseModel):
    submission_id: str
    raw: SnapshotRawData
    parsed: SnapshotParsedData
    resolved: SnapshotResolvedData
    ops: SnapshotOpsData
    errors: SnapshotErrorsData
```
Then, the route can declare the response model using whatever top-level shape the service already returns. For example, if the service returns a bare list, use:

```python
response_model=list[ReviewerSubmissionSnapshot]
```

If the service returns a wrapped object, use a top-level model such as:

```python
class DashboardSnapshotResponse(BaseModel):
    submissions: list[ReviewerSubmissionSnapshot]
```

The implementation should follow the current returned shape from PR #200 rather than changing the payload to fit a new model. The important part is that the returned shape is explicit, typed, and validated.

## Testing expectations
Add or update focused tests to confirm:
* `GET /snapshot` declares and uses the response model.
* The response body matches the expected typed shape.
* Required fields are present.
* Optional fields behave intentionally when source data is missing.
* Existing route and service tests still pass.

A good test here should catch accidental response-shape drift, not merely confirm that the endpoint returns HTTP 200.

## ✅ Acceptance criteria
- [ ] Snapshot response schema is represented with typed Pydantic models.
- [ ] `GET /snapshot` explicitly declares the `response_model` in its route decorator.
- [ ] The response model matches the existing reviewer-facing snapshot contract.
- [ ] Existing route/service tests still pass.
- [ ] Add or update at least one test that confirms the response shape matches the new model.
- [ ] Optional/missing fields are represented intentionally in the model.
- [ ] The route handler remains thin and does not absorb composition logic.

## 🚫 Non-goals
* Changing the underlying snapshot composition behavior.
* Adding auth/gating.
* Adding pagination.
* Adding frontend implementation.
* Changing Sheets/Drive persistence behavior.
* Adding new reviewer workflow actions.

**One-sentence PR rule:** Add typed Pydantic response models to make `GET /snapshot` a stable v0.3 frontend/backend contract *without* changing snapshot composition behavior.

jliu233:
## Summary

Adds explicit typed Pydantic response models for the reviewer-facing `GET /snapshot` API contract.

This makes the current dashboard snapshot shape explicit without changing snapshot composition behavior.

## Changes

- Added typed snapshot response models for:
  - raw submission data
  - parser data
  - resolver data
  - ops state
  - error summary
  - top-level reviewer submission snapshot
- Updated `GET /snapshot` to declare:

```python
response_model=list[ReviewerSubmissionSnapshot]
```

- Added focused route tests that confirm:
  - the route declares the expected response model
  - a valid composed snapshot payload is returned unchanged
  - missing required fields are rejected by FastAPI response validation

## Scope

This PR does not change:

- snapshot composition logic
- auth or gating
- pagination
- persistence behavior
- frontend code
- reviewer workflow behavior

## Testing

`pytest` could not be run locally because `pytest` is not installed in `backend/venv`.

Manual verification with FastAPI `TestClient` passed:

- valid typed `/snapshot` payload returns `200`
- missing required nested fields returns `500`
- route reports `list[api.models.dashboard.ReviewerSubmissionSnapshot]`

Closes #201.

### DIFF
--- a/backend/api/models/__init__.py
+++ b/backend/api/models/__init__.py
@@ -1,0 +1,1 @@
+"""API response models."""
--- a/backend/api/models/dashboard.py
+++ b/backend/api/models/dashboard.py
@@ -1,0 +1,85 @@
+from typing import Literal
+
+from pydantic import BaseModel, ConfigDict
+
+from ops_schema import (
+    OPS_STATUS_CLOSED,
+    OPS_STATUS_CONTACTED,
+    OPS_STATUS_IN_PROGRESS,
+    OPS_STATUS_NEW,
+    OPS_STATUS_PAUSED,
+    OPS_STATUS_REVIEWED,
+)
+
+
+class SnapshotModel(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+
+class SnapshotRawData(SnapshotModel):
+    created_at: str
+    full_name: str
+    email: str
+    location_raw: str
+    timezone: str
+    skills_raw: str
+    interests: str
+    experience_level: str
+    availability: str
+    motivation: str
+    linkedin_url: str
+    github_url: str
+    consent_given: str
+    resume_filename: str
+    resume_status: str
+
+
+class SnapshotParsedData(SnapshotModel):
+    parser_state: Literal["pending", "complete"]
+    parser_run_id: str
+    created_at: str
+    parser_version: str
+    parsed_skills_raw: str
+    parsed_location_raw: str
+    parser_confidence: str
+
+
+class SnapshotResolvedData(SnapshotModel):
+    resolver_state: Literal["not_run", "resolved", "zero_matches"]
+    resolver_version: str
+    aliases_version: str
+    resolved_skill_ids: str
+    unknown_skills: str
+    resolver_coverage: str
+
+
+class SnapshotOpsData(SnapshotModel):
+    status: Literal[
+        OPS_STATUS_NEW,
+        OPS_STATUS_REVIEWED,
+        OPS_STATUS_CONTACTED,
+        OPS_STATUS_IN_PROGRESS,
+        OPS_STATUS_PAUSED,
+        OPS_STATUS_CLOSED,
+    ]
+    notes: str
+    tags: str
+    contact_tracking: str
+    updated_at: str
+    updated_by: str
+
+
+class SnapshotErrorsData(SnapshotModel):
+    has_error: bool
+    latest_error_summary: str
+    latest_error_stage: str
+    latest_error_code: str
+
+
+class ReviewerSubmissionSnapshot(SnapshotModel):
+    submission_id: str
+    raw: SnapshotRawData
+    parsed: SnapshotParsedData
+    resolved: SnapshotResolvedData
+    ops: SnapshotOpsData
+    errors: SnapshotErrorsData
--- a/backend/api/models/dashboard.py
+++ b/backend/api/models/dashboard.py
@@ -2,14 +2,7 @@ from typing import Literal
 
 from pydantic import BaseModel, ConfigDict
 
-from ops_schema import (
-    OPS_STATUS_CLOSED,
-    OPS_STATUS_CONTACTED,
-    OPS_STATUS_IN_PROGRESS,
-    OPS_STATUS_NEW,
-    OPS_STATUS_PAUSED,
-    OPS_STATUS_REVIEWED,
-)
+from ops_schema import VALID_OPS_STATUSES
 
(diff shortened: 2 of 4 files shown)


Kevin-TCUS (approved):
@jliu233 , this looks really good. Approved and ready to merge! 

This is exactly the contract-hardening follow-up I was hoping for after #200. The route stays thin, the response model is declared directly in the `GET /snapshot` decorator, and the new model file makes the reviewer-facing snapshot shape explicit without changing composition behavior.

A few things that stand out here:
1. **Clean separation:** `api.models.dashboard` owns the response contract, `api.routes.dashboard` only wires the route, and `dashboard_service` remains responsible for composition.
2. **Defensive model config:** Adding `ConfigDict(extra="forbid")` is a brilliant API-safety choice. It means the snapshot response will strictly reject undeclared fields instead of silently allowing the contract to drift.
3. **Useful negative test:** The missing-required-fields test using `raise_server_exceptions=False` is a textbook way to verify that FastAPI/Pydantic response validation is actually protecting the endpoint, not just checking that `/snapshot` returns 200.

**One tiny non-blocking note for the future:** The `Literal[VALID_OPS_STATUSES]` pattern depends on `VALID_OPS_STATUSES` being shaped in a way that works cleanly with Pydantic and static type tooling. CI passing suggests it is working perfectly at runtime, but static type checkers (like mypy) sometimes get cranky unless the tuple is explicitly unpacked. Just something to keep in mind if the ops constants are ever refactored!

This gives the frontend work a much more stable, rock-solid contract to build against. Great work, good to merge!

jliu233:
Thanks so much for the thoughtful review!

I’m glad this matches the contract-hardening direction after #200. I’ll keep the `Literal[VALID_OPS_STATUSES]` note in mind, especially if we add stricter static type checking or touch the ops constants later. Since CI is passing and the runtime validation is covered, I’ll leave it as-is for this PR.

Appreciate the approval:)

PR #203, opened 2026-05-07
Render Inbox submission list rows (for issue 176)

### LINKED ISSUE #176: [Frontend] Render Inbox submission list rows from the snapshot contract for 164
### Title: `[Frontend] Render Inbox submission list rows from the snapshot contract`

**🎯 Summary**
Implement the left-side Inbox queue/list so reviewers can scan submissions from the snapshot payload.

**🎯 What we’re doing**
We are rendering the list-view portion of Inbox using only the fields intended for reviewer scanning.
Each row should make it easy to scan: name, location, top skills or signals, workflow status, submitted date, and a lightweight attention/error signal if relevant.

* This ticket is about list-row rendering only.
* It must **not**: implement detail-panel logic, implement writeback, implement advanced filtering, or invent frontend-only fields outside the snapshot contract.
* **Important UI rule:** Each row must use `submission_id` as the stable React key (or equivalent), not array index.

**🛠️ What you need to do**
* Render one list row per snapshot record.
* Use only fields intended for Inbox/list view.
* Keep rows visually scannable.
* Avoid overloading the list with detail-only fields.

**✅ How to tell it’s done**
* [ ] Inbox list renders snapshot-backed records.
* [ ] Each row is tied to one `submission_id`.
* [ ] List-view fields come only from the snapshot contract.
* [ ] Rows use `submission_id` as stable identity/key.
* [ ] Rows are visually scannable and do not overload the list with detail-only fields.

**🚫 Non-goals**
* Detail-panel rendering.
* Writeback.
* Advanced filtering.
* API changes.

---
*One-sentence PR rule: Implements Inbox list-row rendering from the snapshot contract using submission_id as stable row identity.*

jliu233:
## Summary
This PR implements the Inbox submission list-row rendering for reviewer triage using the snapshot contract from #161.

It adds a frontend Inbox view that fetches the snapshot payload and renders one row per submission. Each row is intentionally lightweight and scan-friendly, showing only list-view fields reviewers need for queue triage: applicant name, location, top skills/signals, workflow status, submitted date, and a small attention indicator when the snapshot reports an error.

## Changes
- Added typed frontend snapshot interfaces matching the reviewer snapshot structure.
- Added an `/inbox` route for the reviewer Inbox list.
- Added `InboxSubmissionRow` for rendering snapshot-backed list rows.
- Fetches records from `/snapshot`.
- Uses `submission.submission_id` as the stable React key for each rendered row.
- Displays:
  - `raw.full_name`
  - `parsed.parsed_location_raw` with fallback to `raw.location_raw`
  - `parsed.parsed_skills_raw` with fallback to `raw.skills_raw`
  - `ops.status`
  - `raw.created_at`
  - `errors.has_error` as a lightweight attention signal
- Added Vite proxy support for `/snapshot` during local frontend development.

## Scope Boundaries
This PR is intentionally limited to issue #176.

It does not implement:
- Detail-panel rendering
- Reviewer writeback
- Advanced filtering
- Sorting controls
- API changes
- Frontend-only snapshot fields outside the contract

## Contract Notes
The Inbox rows are rendered directly from the snapshot shape and avoid inventing additional data fields. Row identity is tied to `submission_id`, satisfying the stable-key requirement for the queue list.

## Verification
- Ran `npm run build` successfully.

Closes #176.


Kevin-TCUS (changes requested):
@jliu233, this is a fantastic first step for the Inbox lane. The PR stays beautifully aligned with #176: it renders snapshot-backed list rows, uses submission_id as the stable key, keeps the rows scannable, and avoids drifting into detail-panel logic or writeback.

A few specific things you did really well:

1. State handling: Using a typed InboxState with clear status values makes the loading / ready / error paths much easier to reason about.
2. Smart fallbacks: Falling back to raw location/skills when parsed data is missing is exactly how we want the UI to handle incomplete records.
3. Clean separation: The split between Inbox.tsx, InboxSubmissionRow.tsx, and the snapshot types file is clean and easy to follow.

I have one small requested change before merge:

In Inbox.tsx, after response.json(), could you add a lightweight runtime guard that confirms the payload is actually an array before setting state?

Right now we do:
```typescript
const submissions = (await response.json()) as ReviewerSubmissionSnapshot[]
```

That TypeScript cast helps us while coding, but it does not validate the real API response in the browser. So if /snapshot accidentally returned an error object or a wrapped object instead of a list, the Inbox could crash later when it tries to call .map().

A small guard would turn that into a clean error state instead of a confusing frontend crash. Something like:
```typescript
const data = await response.json()

if (!Array.isArray(data)) {
  throw new Error('Snapshot response must be an array')
}

setState({ status: 'ready', submissions: data as ReviewerSubmissionSnapshot[] })
```
After that, I think this is good to merge. Great job keeping this atomic and making the first Inbox surface feel like a real reviewer queue rather than a spreadsheet clone.

Kevin-TCUS (approved):
@jliu233 , thank you for making the runtime guard change. This addresses the only concern I had.

The Inbox now checks that /snapshot actually returns an array before trying to render rows, which gives us a clean error state instead of a confusing frontend crash if the API shape is ever wrong.

Everything else remains nicely scoped to #176: snapshot-backed list rows, submission_id as the stable key, scannable row content, and no drift into detail panel, writeback, filtering, or sorting.

Approved and good to merge. Great first Inbox UI slice.

Issue #215, opened 2026-05-14
[Frontend] Refactor Inbox detail panel components into separate files without behavior changes


Kevin-TCUS:
### Title: [Frontend] Refactor Inbox detail panel components into separate files without behavior changes

*🔗 Relationship to prior work*
Now that the Inbox detail panel includes raw submission data, parsed output, resolved output, workflow state, and lightweight error visibility, `frontend/src/pages/Inbox.tsx` has grown to nearly 600 lines. This issue is a pure refactor to reduce file complexity before ops writeback controls are added in #169.

*🎯 Summary*
The goal is to preserve all current behavior, UI, and data flow exactly as it is, while improving maintainability by extracting the detail panel sections and UI primitives out of the main page file.

*🎯 What we’re doing*
We are moving the functional components currently living at the bottom of `Inbox.tsx` into smaller, dedicated component files (e.g., inside `frontend/src/components/inbox/` or a similar sensible structure).

⚠️ *Implementation Boundaries*
* **Pure Refactor:** This is an extraction only. Do not change logic, data flow, snapshot interpretation, ordering, filtering, selection behavior, or API calls.
* **No Scope Creep:** Do not introduce new shared UI abstractions outside the Inbox area unless duplication already exists. This is not a design-system refactor.

*🛠️ What you need to do*
* Extract the section components out of `Inbox.tsx` (`InboxDetailPanel`, `WorkflowSection`, `RawSubmissionSection`, `ParsedOutputSection`, `ResolvedOutputSection`).
* Extract the UI primitives (`SectionHeader`, `DetailField`, `ListDetailField`, `StateCallout`, `LinkDetailField`).
* Update imports so `Inbox.tsx` remains completely functional but is significantly reduced in size.

*✅ How to tell it’s done*
- [ ] `Inbox.tsx` is significantly reduced in size.
- [ ] Inbox renders identically before and after the refactor.
- [ ] Existing loading, empty, error, and selected-record states still work.
- [ ] Raw / parsed / resolved / workflow / error separation remains perfectly intact.
- [ ] No snapshot contract changes were made.
- [ ] No API or behavior changes were made.
- [ ] Frontend production build passes.

*🚫 Non-goals*
* Status or notes writeback.
* Filtering or sorting improvements.
* Parser/resolver changes or snapshot shape changes.
* Cross-page UI library refactors or new design system work.
* New reviewer workflow behavior.

[...]

PR #216, opened 2026-05-15
Add lightweight Inbox filters and search over snapshot-backed records (for issue 181)

### LINKED ISSUE #181: [Frontend] Add lightweight Inbox filters and search over snapshot-backed records
### Title: `[Frontend] Add lightweight Inbox filters and search over snapshot-backed records`

**🎯 Summary**
Implement the minimum filtering/search behavior needed to make Inbox useful for reviewer triage.

**🎯 What we’re doing**
We are adding lightweight reviewer-facing filtering/search over the snapshot payload so reviewers can find and triage records more efficiently.

* Filtering/search must be implemented **client-side** over the snapshot payload.
* Do **not**: introduce new backend queries, introduce new search endpoints, or recreate backend composition logic in the frontend.

**🛠️ What you need to do**
* Provide lightweight filtering/search for fields such as: skills / keywords, workflow status, location, or other basic reviewer-friendly fields already exposed by the snapshot contract.

**✅ How to tell it’s done**
* [ ] Reviewers can search/filter snapshot-backed Inbox records.
* [ ] Filtering/search is client-side only.
* [ ] Filtering/search is lightweight and reviewer-friendly.
* [ ] No new backend queries are introduced for this ticket.

**🚫 Non-goals**
* Advanced search engines.
* Analytics.
* New backend endpoints or backend filtering logic.

---
*One-sentence PR rule: Implements lightweight client-side Inbox filtering and search over snapshot-backed records without introducing new backend queries.*

jliu233:
## Summary

Implements lightweight client-side filtering and search for the reviewer Inbox, backed entirely by the existing `/snapshot` payload.

Closes #181.

## What Changed

- Added a reviewer-friendly search box to the Inbox list.
- Added workflow status filtering using the existing snapshot `ops.status` values.
- Added a location filter that checks parsed location first, then raw submitted location.
- Added a clear filters action.
- Added visible result counts, showing how many submissions match the current filters.
- Added an empty filtered-state message when records exist but none match.
- Updated selected submission behavior so the detail panel only shows a record that is still present in the filtered results.

## Implementation Notes

Filtering/search is performed entirely in `frontend/src/pages/Inbox.tsx` against the already-loaded snapshot records.

Search includes lightweight reviewer-facing fields already exposed by the snapshot contract, including:

- Submission ID
- Name and email
- Raw and parsed skills
- Raw and parsed location
- Interests, availability, motivation, experience level
- Resume metadata
- Parser and resolver states
- Resolved and unknown skills
- Workflow status, tags, notes, contact tracking
- Latest error summary, stage, and code

No backend queries, endpoints, or backend filtering logic were introduced.

## Validation

- Ran `npm run build`
- Build passed successfully

## PR Rule

Implements lightweight client-side Inbox filtering and search over snapshot-backed records without introducing new backend queries.

### DIFF
--- a/frontend/src/pages/Inbox.tsx
+++ b/frontend/src/pages/Inbox.tsx
@@ -251,40 +251,49 @@ function getInboxSearchText(submission: ReviewerSubmissionSnapshot) {
   return normalizeFilterText(
     [
       submission.submission_id,
-      submission.raw.full_name,
-      submission.raw.email,
-      submission.raw.location_raw,
-      submission.raw.timezone,
-      submission.raw.skills_raw,
-      submission.raw.interests,
-      submission.raw.experience_level,
-      submission.raw.availability,
-      submission.raw.motivation,
-      submission.raw.resume_filename,
-      submission.raw.resume_status,
-      submission.parsed.parser_state,
-      submission.parsed.parsed_skills_raw,
-      submission.parsed.parsed_location_raw,
-      submission.resolved.resolver_state,
-      submission.resolved.resolved_skill_ids,
-      submission.resolved.unknown_skills,
-      submission.ops.status,
-      submission.ops.notes,
-      submission.ops.tags,
-      submission.ops.contact_tracking,
-      submission.errors.latest_error_summary,
-      submission.errors.latest_error_stage,
-      submission.errors.latest_error_code
-    ].join(' ')
+      submission.raw?.full_name,
+      submission.raw?.email,
+      submission.raw?.location_raw,
+      submission.raw?.timezone,
+      submission.raw?.skills_raw,
+      submission.raw?.interests,
+      submission.raw?.experience_level,
+      submission.raw?.availability,
+      submission.raw?.motivation,
+      submission.raw?.resume_filename,
+      submission.raw?.resume_status,
+      submission.parsed?.parser_state,
+      submission.parsed?.parsed_skills_raw,
+      submission.parsed?.parsed_location_raw,
+      submission.resolved?.resolver_state,
+      submission.resolved?.resolved_skill_ids,
+      submission.resolved?.unknown_skills,
+      submission.ops?.status,
+      submission.ops?.notes,
+      submission.ops?.tags,
+      submission.ops?.contact_tracking,
+      submission.errors?.latest_error_summary,
+      submission.errors?.latest_error_stage,
+      submission.errors?.latest_error_code
+    ]
+      .map(safeText)
+      .join(' ')
   )
 }
 
 function getPreferredLocation(submission: ReviewerSubmissionSnapshot) {
-  return submission.parsed.parsed_location_raw.trim() || submission.raw.location_raw.trim()
+  return (
+    safeText(submission.parsed?.parsed_location_raw).trim() ||
+    safeText(submission.raw?.location_raw).trim()
+  )
 }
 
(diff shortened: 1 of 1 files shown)


Kevin-TCUS (changes requested):
Jiaqi, this is a strong implementation of #181. The overall approach looks right: filters are client-side over the loaded `/snapshot` records, no backend query/API changes were introduced, and the UX additions like clear filters, result count, and filtered-empty state are exactly the right scope.

I do want to request one small defensive hardening change before merge.

Because the search/filter helpers touch many snapshot fields, they should be extra safe around missing or unexpected API values. The snapshot contract expects strings, but API data is still runtime data, and a malformed or partial response could crash the UI if we call `.trim()` on something null or undefined.

The main place I noticed this is:
```typescript
function getPreferredLocation(submission: ReviewerSubmissionSnapshot) {
  return submission.parsed.parsed_location_raw.trim() || submission.raw.location_raw.trim()
}
```

Could you make this null-safe before trimming? For example:
```

function safeText(value: unknown) {
  return typeof value === 'string' ? value : ''
}

function getPreferredLocation(submission: ReviewerSubmissionSnapshot) {
  return (
    safeText(submission.parsed?.parsed_location_raw).trim() ||
    safeText(submission.raw?.location_raw).trim()
  )
}
```
It would also be good to use the same pattern in getInboxSearchText, since that function reads many nested raw / parsed / resolved / ops / error fields. The goal is not to change the filtering behavior, just to make the helpers resilient if a snapshot field is unexpectedly missing.

Once that defensive guard is in place, this looks good to merge. Great work adding the first real reviewer utility layer to the Inbox!

Kevin-TCUS (changes requested):
**Review Status:** Request Changes

**Comment:**

Jiaqi, thank you for the quick update. This is very close.

The new `safeText()` helper and the updates to `getPreferredLocation`, `normalizeFilterText`, and `getInboxSearchText` address the main runtime-safety issue I was worried about.

One tiny follow-up before I approve: in `matchesInboxFilters`, this line still assumes `submission.ops` always exists:

```typescript
if (filters.statusFilter !== 'all' && submission.ops.status !== filters.statusFilter) {
  return false
}
```

Could you make that one status check defensive too? For example:

```typescript
const submissionStatus = safeText(submission.ops?.status)

if (filters.statusFilter !== 'all' && submissionStatus !== filters.statusFilter) {
  return false
}
```

After that, I think this is good to merge. The filter/search implementation itself looks strong and stays nicely inside the #181 scope.

Kevin-TCUS (approved):
@jliu233 , this looks good now. Approved.

Thank you for making the filter/search helpers defensive. The new `safeText()` helper, optional chaining in `getInboxSearchText`, the null-safe `getPreferredLocation`, and the defensive status check in `matchesInboxFilters` address the runtime-safety issue I was concerned about.

This PR stays nicely inside #181:

- filtering/search remains client-side over the loaded `/snapshot` records
- no backend queries or API changes were introduced
- status filtering uses the existing `OpsStatus` contract
- location filtering uses parsed location first, then raw location
- result counts and filtered-empty state make the UI behavior clear
- selected detail behavior is reasonable for the filtered result set

There are still broader opportunities to break up and harden `Inbox.tsx`, but that is already covered by #215 and should not block this PR.

Great work. This adds a useful reviewer utility layer without expanding the backend scope.

PR #218, opened 2026-05-19
Implements default actionable ordering for the Inbox queue for issue 182

### LINKED ISSUE #182: [Frontend] Implement default actionable ordering for the Inbox queue
### Title: `[Frontend] Implement default actionable ordering for the Inbox queue`

**🎯 Summary**
Ensure the Inbox queue defaults to a useful reviewer-first order rather than an arbitrary list.

**🎯 What we’re doing**
We are implementing default actionable ordering so reviewers see the most important work first.

* Default actionable ordering is applied in the **frontend** over the snapshot payload.
* Do **not**: move this into backend query logic for this ticket, create a complex prioritization engine, or create inconsistent ordering behavior relative to the Inbox product model.

**🛠️ What you need to do**
* Implement a default ordering that prioritizes actionable submissions, such as: `new`, `unreviewed`, requiring attention, or blocked states if surfaced as attention-worthy.

**✅ How to tell it’s done**
* [ ] Inbox has a defined default order.
* [ ] Actionable work is surfaced before less urgent records.
* [ ] Ordering behavior is applied in the frontend over the snapshot payload.
* [ ] Ordering is consistent and understandable.

**🚫 Non-goals**
* Complex prioritization engine.
* Backend sorting logic for this ticket.
* Analytics-driven ranking.

---
*One-sentence PR rule: Implements default actionable ordering in the frontend so Inbox surfaces the most review-relevant submissions first.*

jliu233:
## Summary

The reviewer Inbox now sorts the snapshot payload on the frontend so the most review-relevant submissions appear first by default, instead of relying on arbitrary backend or sheet order.

## Related Issue

Closes #182

## Changes

- Added frontend-only queue ordering in `frontend/src/pages/Inbox.tsx`
- Applies ordering over the existing `/snapshot` payload before search, status, and location filters
- Keeps the fetched snapshot array immutable by sorting a copied array
- Adds a deterministic comparator for Inbox submissions
- Uses only fields from the reviewer-facing snapshot contract:
  - `errors.has_error`
  - `resolved.resolver_state`
  - `resolved.unknown_skills`
  - `ops.status`
  - `raw.created_at`
  - `submission_id`

## Default Ordering

The Inbox now prioritizes submissions in this order:

1. Submissions with active error signals
2. Submissions with resolver attention signals:
   - `resolver_state === "zero_matches"`
   - non-empty `unknown_skills`
3. `new` submissions
4. `paused` submissions
5. `in_progress` submissions
6. `contacted` submissions
7. `reviewed` submissions
8. `closed` submissions

Within the same priority group, newer submissions are shown first using `raw.created_at`.

If timestamps are equal or unavailable, `submission_id` is used as a final tie-breaker so ordering remains stable and understandable.

## Implementation Notes

This intentionally stays within the issue boundaries:

- No backend query sorting was added
- No prioritization engine was introduced
- No snapshot contract changes were made
- Existing filters still work the same way
- Ordering is applied in the frontend over the snapshot payload

## Testing

Ran:

```bash
npm run build
```

Result:

```text
✓ built successfully
```

### DIFF
--- a/frontend/src/pages/Inbox.tsx
+++ b/frontend/src/pages/Inbox.tsx
@@ -18,13 +18,15 @@ export default function Inbox() {
   const filteredSubmissions = useMemo(() => {
     if (state.status !== 'ready') return []
 
-    return state.submissions.filter((submission) =>
-      matchesInboxFilters(submission, {
-        searchQuery,
-        statusFilter,
-        locationFilter
-      })
-    )
+    return [...state.submissions]
+      .sort(compareInboxSubmissions)
+      .filter((submission) =>
+        matchesInboxFilters(submission, {
+          searchQuery,
+          statusFilter,
+          locationFilter
+        })
+      )
   }, [locationFilter, searchQuery, state, statusFilter])
 
   const selectedSubmission = useMemo(() => {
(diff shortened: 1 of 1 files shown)


Kevin-TCUS (changes requested):
@jliu233 , this is very close and the overall approach is right. I like that the ordering is frontend-only, uses only snapshot-backed fields, uses deterministic tie-breakers, and keeps the existing filtering behavior unchanged.

I do need one small logic fix before approving, plus one simple optimization.

**1. Rank overlap bug**

The PR description says the Inbox should prioritize:
1. active error signals
2. resolver attention signals
3. `new`
4. `paused`
5. `in_progress`
6. `contacted`
7. `reviewed`
8. `closed`

But the implementation currently gives errors the same rank as `new`, and resolver attention the same rank as `paused`:

```typescript
if (submission.errors?.has_error) return 0
if (hasResolverAttentionSignal(submission)) return 1

const actionableStatusRank: Record<OpsStatus, number> = {
  new: 0,
  paused: 1,
  in_progress: 2,
  contacted: 3,
  reviewed: 4,
  closed: 5
}
```

That means error records can be mixed with `new` records by date, and resolver-attention records can be mixed with `paused` records by date, rather than always appearing before them.

Could you offset the status ranks so the categories are distinct? For example:

```typescript
const actionableStatusRank: Record<OpsStatus, number> = {
  new: 2,
  paused: 3,
  in_progress: 4,
  contacted: 5,
  reviewed: 6,
  closed: 7
}
```

Then:

```typescript
if (submission.errors?.has_error) return 0
if (hasResolverAttentionSignal(submission)) return 1
```

After that, the comparator should match the stated priority order.

**2. Filter before sort**

Right now the code sorts the full snapshot array and then filters:

```typescript
return [...state.submissions]
  .sort(compareInboxSubmissions)
  .filter((submission) => matchesInboxFilters(submission, { ... }))
```

Since `.filter()` already returns a new array, we can filter first and then sort only the visible subset:

```typescript
return state.submissions
  .filter((submission) =>
    matchesInboxFilters(submission, {
      searchQuery,
      statusFilter,
      locationFilter
    })
  )
  .sort(compareInboxSubmissions)
```

That avoids sorting submissions that will not be shown, while still avoiding mutation of `state.submissions`.

**Hardening note**

One tiny related hardening note: `getSortableDateValue` could accept `unknown` or `string | undefined`, since it is being called with `first.raw?.created_at`. Not critical if the build passes, but it would align with the defensive style we just added in #216.

Once the rank values are corrected, I think this is good to merge. The structure is nicely scoped for #182.

Kevin-TCUS (approved):
@jliu233 , this looks good now. Approved.

Thank you for making the requested updates. The rank values now match the intended priority order:

1. active error signals
2. resolver attention signals
3. new
4. paused
5. in_progress
6. contacted
7. reviewed
8. closed

The filter-before-sort change also looks good. Since .filter() returns a new array, the sort is still non-mutating while avoiding sorting records that will not be shown.

I also appreciate the getSortableDateValue(value: unknown) hardening. That aligns nicely with the defensive style added in #216.

This stays well scoped to #182: frontend-only ordering, no backend query changes, no prioritization engine, and no snapshot contract changes.

Great work.

Issue #169, opened 2026-04-21
[API/UI] Implement ops status and notes writeback for the v0.3 reviewer workflow


Kevin-TCUS:
### Title: `[API/UI] Implement ops status and notes writeback for the v0.3 reviewer workflow`

**🔗 Relationship to prior work**
This ticket is the primary mutable workflow step under the `[Epic] Reviewer Dashboard MVP over the v0.3 4-tab schema`.
It builds directly on:
* `Libelle v0.3 Milestone Brief, Architecture, and Execution Map` #76
* `Add repo-owned ops status constants and validation helper` #135
* `[API] Define the reviewer-facing GET /snapshot contract for the v0.3 dashboard` #161
* `[Frontend] Build Inbox master-detail view for reviewer triage` #164

It also aligns with:
* `[Security] Gate deployed dashboard access and derive actor identity for internal actions` #165

The dashboard needs this write path so reviewers can triage submissions without editing raw Google Sheets directly. For deployed/internal use, actor-aware write behavior should align with the security/access ticket.

**🎯 Summary**
Implement the minimum write path so reviewers can update a submission’s status and notes from the dashboard and persist that state into the `ops` tab.
In v0.3, `ops` is the mutable human workflow layer. Its job is to capture the reviewer’s current decision state for a given `submission_id`.

**🎯 What we’re doing**
We are building the minimum backend + frontend write path for reviewer workflow state.
The write path must:
* target the `ops` tab in Google Sheets
* validate status values against the repo-owned ops status contract
* record actor and timestamp on every write
* allow the dashboard to reflect the updated workflow state smoothly

**🧠 Required v0.3 Semantics**
* **Keying rule:** All `ops` writes are keyed strictly by `submission_id`.
* **Create-vs-update rule:** If no `ops` row exists for a given `submission_id`, the write path must create a new row. If an `ops` row already exists, the write path updates the current workflow state in place.
* **Write rule:** v0.3 uses last-write-wins semantics for mutable ops state. Unlike `parser_results` and `errors`, `ops` is not append-only in v0.3.
* **Actor/timestamp rule:** Every `ops` write must record: `submission_id`, actor identity, timestamp, and updated workflow fields.
* **Status validation rule:** The backend must validate incoming status values using the repo-owned ops contract. The frontend must not invent freeform statuses.
[...]

Issue #165, opened 2026-04-21
[Security] Gate deployed dashboard access and derive actor identity for internal actions


Kevin-TCUS:
### Title: `[Security] Gate deployed dashboard access and derive actor identity for internal actions`

**🔗 Relationship to prior work**
This ticket is the internal-access step under the `[Epic] Reviewer Dashboard MVP over the v0.3 4-tab schema`.
It builds directly on:
* `Libelle v0.3 Milestone Brief, Architecture, and Execution Map` #76
* `Refactor backend toward the v0.3 modular architecture` #74
* `[API] Define the reviewer-facing GET /snapshot contract for the v0.3 dashboard`
* `[API] Implement GET /snapshot over submissions, parser_results, ops, and lightweight error visibility`
* `[Frontend] Build the v0.3 dashboard shell over the 4-tab schema`
* `[Frontend] Build Inbox master-detail view for reviewer triage`
* `[API/UI] Implement ops status and notes writeback for the v0.3 reviewer workflow`

With the dashboard now acting as an internal reviewer decision surface, the next requirement is to ensure that deployed access is gated appropriately and that workflow writes are tied to a real internal actor.

**🎯 Summary**
Put the deployed dashboard behind Cloudflare Access and ensure authenticated identity is available to the backend for internal write actions.
This ticket is about internal access control for v0.3. It is not about building a custom user-management system, role hierarchy, or external-facing login flow.

**🎯 What we’re doing**
We are adding the minimum security and identity layer required for the v0.3 internal dashboard.
For v0.3:
* the dashboard is an internal tool
* public users do not access it directly
* reviewer actions must be attributable to a real actor
* backend write paths must know who performed the action

The goal is to make the dashboard safely usable by internal reviewers without introducing a brand-new auth product.

**🧠 v0.3 Access Model Reminder**
For v0.3, internal dashboard access is expected to be gated by Cloudflare Access.
That means:
* the deployed dashboard is not publicly accessible
* authenticated identity is supplied to the backend through Cloudflare-provided headers and/or JWT context
* backend writes use that identity for actor-aware workflow updates

This ticket should keep that model simple and explicit.

**📋 What this ticket must cover**

* **Dashboard gating:** The deployed dashboard must be behind Cloudflare Access or the equivalent configured internal-access mechanism used for Libelle. A reviewer should not be able to access the deployed dashboard anonymously.
[...]

PR #223, opened 2026-05-26
Create a new ops row when workflow state is first written for a submission (for issue 184)

### LINKED ISSUE #184: [API] Create a new ops row when workflow state is first written for a submission for 169
### Title: `[API] Create a new ops row when workflow state is first written for a submission`

This is an atomic sub-issue under `#169` (`[API/UI] Implement ops status and notes writeback for the v0.3 reviewer workflow`).
It depends directly on:
* `Add repo-owned ops status constants and validation helper` #135
* `[API] Define the reviewer-facing GET /snapshot contract for the v0.3 dashboard` #161

**🎯 Summary**
Implement the create-if-missing path for `ops` writeback so a submission with no existing `ops` row can receive its first workflow state.

**🎯 What we’re doing**
We are implementing the “first write” behavior for the mutable `ops` layer. In v0.3, `ops` is not append-only. It is a mutable current-state layer with one row per `submission_id`. That means if no `ops` row exists yet, the backend must create one rather than failing or assuming the row already exists.

* This ticket is about create-if-missing behavior **only**.
* It must **not**: implement update-in-place behavior for existing `ops` rows, implement frontend UI wiring, implement security/actor derivation policy beyond consuming the actor provided by the trusted write path, or introduce append-only `ops` history.
* All persistence must remain aligned to the repo-owned `ops` schema.
* **Concurrency / duplicate-row guardrail:** The implementation should avoid creating duplicate `ops` rows for the same `submission_id` under concurrent writes (e.g., by re-checking existence before insert or enforcing idempotent behavior at the storage layer where feasible).

**🛠️ What you need to do**
* Detect when no `ops` row exists for a given `submission_id`.
* Create a new `ops` row with the provided workflow fields.
* Persist actor and timestamp metadata on creation.
* Preserve one-row-per-submission semantics for `ops` in v0.3.

**✅ How to tell it’s done**
* [ ] If no `ops` row exists, the backend creates one.
* [ ] The new row is keyed by `submission_id`.
* [ ] Actor and timestamp are captured on creation.
* [ ] The created row aligns to the repo-owned `ops` schema.
* [ ] The create path does not naively create duplicate rows when concurrent writes are possible.

**🚫 Non-goals**
* Updating an existing `ops` row.
* Frontend implementation.
* Append-only `ops` history.
* Bulk writes.

---
*One-sentence PR rule: Implements create-if-missing behavior for the first ops workflow write on a submission_id.*

jliu233:
## Summary

Implements create-if-missing behavior for the first ops workflow write on a `submission_id`.

This adds the backend path needed for issue #184: when a reviewer workflow state is written for a submission that does not yet have an ops row, the API now creates the initial mutable ops row instead of assuming one already exists.

## Related Issue

Closes #184  
Parent work: #169

## Changes

- Added `POST /submissions/{submission_id}/ops`
  - Accepts the first reviewer-facing ops workflow state for a submission.
  - Validates incoming `status` using the existing repo-owned ops status helper.
  - Returns `201` with the created ops state when a new row is created.
  - Returns `200` with `already_exists` when an ops row already exists.

- Added ops write models
  - `OpsWorkflowStateCreateRequest`
  - `OpsWorkflowStateCreateResponse`

- Added ops write service
  - `create_first_ops_workflow_state(...)`
  - Keeps write orchestration separate from route handling.
  - Intentionally does not implement update-in-place behavior.

- Added Sheets persistence support
  - `create_ops_row_if_missing(...)`
  - Detects whether an ops row already exists for the `submission_id`.
  - Appends a schema-aligned ops row when missing.
  - Captures actor metadata via `updated_by`.
  - Captures timestamp metadata via `updated_at`.
  - Uses the existing repo-owned `ops` sheet schema.

- Added duplicate-row guardrail
  - Uses an in-process write lock.
  - Re-checks existing ops rows inside the locked section before appending.
  - Avoids naive duplicate appends during concurrent writes within the same backend process.

- Added focused tests
  - First write creates a schema-aligned ops row.
  - Existing ops row returns no-op behavior.
  - Route returns created response for first write.
  - Route does not update existing ops row.
  - Invalid ops status is rejected before persistence.

## Non-Goals Preserved

This PR does not:

- Implement update-in-place behavior for existing ops rows.
- Add frontend UI wiring.
- Add append-only ops history.
- Add bulk writes.
- Introduce actor derivation or authorization policy beyond consuming `updated_by` from the trusted write path.

### Create first ops workflow state

`POST /submissions/{submission_id}/ops`

Example request:

```json
{
  "status": "contacted",
  "notes": "Left voicemail",
  "tags": "priority",
  "contact_tracking": "call",
  "updated_by": "reviewer@example.org"
}
```

Example created response:

```json
{
  "status": "created",
  "submission_id": "sub_001",
  "ops": {
    "status": "contacted",
    "notes": "Left voicemail",
    "tags": "priority",
    "contact_tracking": "call",
    "updated_at": "05-26-2026 10:00:00 UTC",
    "updated_by": "reviewer@example.org"
  }
}
```

Example existing-row response:

```json
{
  "status": "already_exists",
  "submission_id": "sub_001",
  "ops": null
}
```

## Validation

Incoming `status` values continue to use the repo-owned ops status contract:

- `new`
- `reviewed`
- `contacted`
- `in_progress`
- `paused`
- `closed`

Invalid statuses return `400` before any persistence call is made.

## Testing

Added tests covering:

- `create_first_ops_workflow_state` appends a schema-aligned row when missing.
- Existing ops rows are not modified or duplicated.
- The write path re-checks existing rows before append.
- The API route returns the expected created response.
- The API route returns no-op behavior for existing rows.
- Invalid status blocks persistence.

Local verification performed:

```bash
python -m py_compile backend/storage/sheets_repo.py backend/services/ops_write_service.py backend/api/routes/dashboard.py backend/api/models/dashboard.py backend/tests/test_ops_write_service.py backend/tests/test_dashboard_route.py
git diff --check
```

Could not run pytest locally because `pytest` is not installed in the active Python environment:

```text
No module named pytest
```

### DIFF
--- a/backend/api/models/dashboard.py
+++ b/backend/api/models/dashboard.py
@@ -55,6 +55,20 @@ class SnapshotOpsData(SnapshotModel):
     updated_by: str
 
 
+class OpsWorkflowStateCreateRequest(SnapshotModel):
+    status: str
+    notes: str = ""
+    tags: str = ""
+    contact_tracking: str = ""
+    updated_by: str
+
+
+class OpsWorkflowStateCreateResponse(SnapshotModel):
+    status: Literal["created", "already_exists"]
+    submission_id: str
+    ops: SnapshotOpsData | None
+
+
 class SnapshotErrorsData(SnapshotModel):
     has_error: bool
     latest_error_summary: str
--- a/backend/api/routes/dashboard.py
+++ b/backend/api/routes/dashboard.py
@@ -1,7 +1,13 @@
-from fastapi import APIRouter
+from fastapi import APIRouter, Response, status as http_status
 
-from api.models.dashboard import ReviewerSubmissionSnapshot
+from api.models.dashboard import (
+    OpsWorkflowStateCreateRequest,
+    OpsWorkflowStateCreateResponse,
+    ReviewerSubmissionSnapshot,
+)
+from api.ops_status_validation import validate_incoming_ops_status
 from services.dashboard_service import get_snapshot_records
+from services.ops_write_service import create_first_ops_workflow_state
 
 router = APIRouter()
 
@@ -9,3 +15,47 @@ router = APIRouter()
 @router.get("/snapshot", response_model=list[ReviewerSubmissionSnapshot])
 def get_snapshot():
     return get_snapshot_records()
+
+
+@router.post(
+    "/submissions/{submission_id}/ops",
+    response_model=OpsWorkflowStateCreateResponse,
+    status_code=201,
+)
+def create_ops_workflow_state(
+    submission_id: str,
+    payload: OpsWorkflowStateCreateRequest,
+    response: Response,
+):
+    status = validate_incoming_ops_status(payload.status)
+    created_row = create_first_ops_workflow_state(
+        submission_id,
+        {
+            "status": status,
+            "notes": payload.notes,
+            "tags": payload.tags,
+            "contact_tracking": payload.contact_tracking,
+            "updated_by": payload.updated_by,
+        },
+    )
+
+    if created_row is None:
+        response.status_code = http_status.HTTP_200_OK
+        return {
+            "status": "already_exists",
+            "submission_id": submission_id,
+            "ops": None,
+        }
+
+    return {
+        "status": "created",
+        "submission_id": submission_id,
+        "ops": {
+            "status": created_row["status"],
+            "notes": created_row["notes"],
+            "tags": created_row["tags"],
+            "contact_tracking": created_row["contact_tracking"],
+            "updated_at": created_row["updated_at"],
+            "updated_by": created_row["updated_by"],
+        },
+    }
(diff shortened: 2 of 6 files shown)


Kevin-TCUS (approved):
@jliu233, this looks good to me. Approved and ready to merge.

This is nicely scoped to #184. It adds the create-if-missing path for the first ops row without trying to handle updates yet, which is exactly the right split for #169.

A few things I checked:

* first write returns 201 with the created ops state
* existing ops row returns 200 / already_exists instead of updating
* invalid status is rejected before persistence
* the write path uses the existing ops schema helper
* update-in-place behavior is correctly left for the next issue

One note for later, not a blocker here: I noticed `updated_by` is currently being accepted directly from the request body. That is completely fine for this PR since we don't have auth wired up yet, but let's keep #165 (Actor Identity) on our radar. Once this becomes a real authenticated workflow, we'll want to derive that identity securely from the session rather than trusting the client payload.

Nice work keeping this atomic and moving the #169 writeback path forward.

Issue #130, opened 2026-04-10
[Task] Add submission_id to core intake and parser logs


Kevin-TCUS:
### 🔗 Relationship to prior work
This is the next atomic implementation step under #70 (Generate and standardize submission_id across the intake pipeline). It builds on the intake-time UUID generation work and improves end-to-end traceability.

### 🎯 Summary
Once `submission_id` exists, the backend logs should use it consistently so a maintainer can trace a single submission across upload, write, parse, and writeback flow.

*This ticket is intentionally narrow:* update core log messages to include `submission_id` where it matters most.

### 🛠️ Implementation Steps
1. **Update intake logs:** Add `submission_id={submission_id}` to the core log lines in the intake / upload flow.
2. **Update parser-related logs:** Add `submission_id={submission_id}` to the async parser / writeback flow wherever that ID is available.
3. **Preserve current behavior:** This is a logging improvement only. Do not redesign parser or intake behavior in this ticket.

### ✅ Definition of Done
- [ ] Core intake log messages include `submission_id`.
- [ ] Core parser / writeback log messages include `submission_id` where available.
- [ ] A maintainer can follow one submission through the main log flow using that ID.
- [ ] *No unrelated refactors:* No unrelated refactors are included.

### 🚫 Non-goals
- Changing parser extraction behavior.
- Refactoring startup validation.
- Redesigning the logging framework.
- Moving code into new service/storage layers.

### 📝 One-Sentence PR Rule
Your PR description for this task should simply be:
*"Adds submission_id to core intake and parser logs for end-to-end traceability."*

**🏷️ Labels:** `backend`, `observability`, `atomic unit`
**🏁 Milestone:** v0.3 – Trustworthy Volunteer Intake

Issue #129, opened 2026-04-10
[Task] Generate UUIDv4 submission_id at intake and return it in the API response


Kevin-TCUS:
### 🔗 Relationship to prior work
This is the next atomic implementation step under #70 (Generate and standardize submission_id across the intake pipeline). It establishes the canonical ID at the very start of the intake flow.

### 🎯 Summary
We need one stable `submission_id` that is created at intake time and returned immediately to the client. This becomes the anchor for the rest of the pipeline.

*This ticket is intentionally narrow:* generate the UUIDv4 ID in the intake route and return it in the success response.

### 🛠️ Implementation Steps
1. **Generate the ID at intake start:** In the upload / intake route, import Python’s `uuid` library and generate a `str(uuid.uuid4())` string as soon as a valid submission is being processed.
2. **Use `submission_id` consistently inside the intake flow:** Ensure the route uses that generated ID for the current request context.
3. **Return it in the API response:** Include `submission_id` in the success JSON returned to the client.

### ✅ Definition of Done
- [ ] A UUIDv4 `submission_id` is generated at the start of the intake request.
- [ ] The same ID is returned in the success response JSON.
- [ ] *No unrelated refactors:* No other systems are refactored in this ticket.

### 🚫 Non-goals
- Passing `submission_id` into the async parser task.
- Refactoring Google Drive naming.
- Refactoring Google Sheets write logic beyond what is needed to preserve current behavior.
- Changing startup validation logic.

### 📝 One-Sentence PR Rule
Your PR description for this task should simply be:
*"Generates UUIDv4 submission_id at intake and returns it in the API response."*

**🏷️ Labels:** `backend`, `api`, `atomic unit`
**🏁 Milestone:** v0.3 – Trustworthy Volunteer Intake

Issue #125, opened 2026-04-10
[Task] Swap submissions write logic to strict append-only


Kevin-TCUS:
### 🔗 Relationship to prior work
This is an atomic implementation step under #71 (Implement append-only write logic for new submissions). It builds on the schema-driven submissions work and helps ensure that raw intake data is never overwritten.

### 🎯 Summary
We need the submissions write path to behave as a true append-only insertion flow. The backend should add a new row at the bottom of the `submissions` tab for each submission, rather than relying on manual row targeting or update-style behavior.

*This ticket is intentionally narrow:* change the submissions write path to use strict append-only Google Sheets behavior.

### 🛠️ Implementation Steps
1. **Locate the submissions write API call:** Find the Google Sheets write call used in the submissions flow (currently in `sheets_sync.py` or the equivalent submissions write path).
2. **Use append-only behavior:** Ensure the backend uses the Google Sheets `.append()` method for the `submissions` tab write path instead of update-style behavior or any manual row-targeting logic.
3. **Remove counter/manual-row assumptions:** Remove any reliance on `RESUME_COUNTER`, manual row calculations, or logic that assumes the backend needs to know the current row position before writing.
4. **Preserve current data behavior:** This ticket should change *how* the submission row is written, not *what* values are written. The submissions row should still contain the same logical data as before.

### ✅ Definition of Done
- [ ] The `submissions` write path uses Google Sheets append-only behavior.
- [ ] The backend successfully appends new rows to the bottom of the `submissions` tab without manually calculating row positions.
- [ ] Any counter-based or manual row-targeting logic is removed from this write path.
- [ ] Submitting multiple applications creates multiple rows without overwriting prior submissions.

### 🚫 Non-goals
- Refactoring `parser_results` writes.
- Changing the `submissions` schema.
- Changing Drive upload behavior.
- Refactoring startup validation logic.

### 📝 One-Sentence PR Rule
Your PR description for this task should simply be:
*"Swaps the submissions write path to strict append-only Google Sheets behavior."*

**🏷️ Labels:** `backend`, `sheets`, `atomic unit`
**🏁 Milestone:** v0.3 – Trustworthy Volunteer Intake

Kevin-TCUS:
Closing as completed.

The current submissions write path uses strict append-only Google Sheets behavior through `values().append(...)` in `write_base_row(...)`.

This satisfies the core requirements of this issue:

- new submissions are appended to the `submissions` tab
[...]

PR #229, opened 2026-05-31
Pass submission_id through the async parser context (#126)

### LINKED ISSUE #126: [Task] Pass submission_id through the async parser context
### 🔗 Relationship to prior work
This is the next atomic implementation step under #70 (Generate and standardize submission_id across the intake pipeline). It builds on the intake-time UUID generation work and prepares the parser flow to stay linked to the same submission across asynchronous execution.

### 🎯 Summary
Now that the intake route generates a `submission_id`, the background parsing task also needs to receive that same ID so the parser flow can stay tied to the correct submission.

*This ticket is intentionally narrow:* update the async parser context so `submission_id` is passed through and logged.

### 🛠️ Implementation Steps
1. **Update the background task signature:** Update the background parsing task function signature (for example, `_parse_and_update`) so it accepts `submission_id: str` as a required parameter.
2. **Pass `submission_id` from the intake route:** In the intake/upload route, when the background task is triggered, pass the newly generated `submission_id` into that task.
3. **Add confirmation logging:** Add a clear log line inside the background task to confirm it received the `submission_id`, for example:
   ```python
   logger.info(f"Starting background parse for submission_id={submission_id}")
   ```
4. **Keep the flow intact:** Ensure this change does not break the main intake response path or the async parser execution flow.

### ✅ Definition of Done
- [ ] The background parsing task accepts `submission_id` as an explicit parameter.
- [ ] The intake route passes the generated `submission_id` into the background task.
- [ ] The background task logs the received `submission_id`.
- [ ] The async parsing task still executes without breaking the main submission response.

### 🚫 Non-goals
- Changing parser extraction behavior.
- Writing `parser_results` via schema helpers in this ticket.
- Changing startup validation logic.
- Refactoring the full async pipeline into a service layer.

### 📝 One-Sentence PR Rule
Your PR description for this task should simply be:
*"Passes submission_id through the async parser context and logs it in the background parsing flow."*

**🏷️ Labels:** `backend`, `api`, `atomic unit`
**🏁 Milestone:** v0.3 – Trustworthy Volunteer Intake

briannadelariva:
## Summary

Closes #126. Part of the v0.3 `submission_id` lane under #74 / #70; follows #129 (merged in #205).

Passes `submission_id` through the async parser context by stamping it onto the parser result payload, so the parser's output is reliably linked back to the original submission.

## What changed

`backend/services/parser_service.parse_and_update`:
- Adds `parsed["submission_id"] = submission_id` alongside the existing `parsed["drive_file_id"] = drive_file_id`. The parser-result payload now self-describes its link to the originating submission — the canonical ID travels with the data, not just via a side-channel arg.
- Drops the default on `pre_extracted_text` so the signature is explicit per the issue's "explicit parameter" language. The intake route already passes it positionally, so no caller change is required.

The log line is intentionally unchanged in shape (`[JOB] Parsing submission_id={submission_id} ...`) — already includes the ID. Structured-logger migration is deferred to #130.

## Why this approach

Background context: PR #154 already plumbed `submission_id` through the function call chain (route → `parse_and_update` → `update_resume_in_sheet`). What was still missing for #126 was *using* the ID inside the parser context itself, not just passing it as an arg. Stamping it onto the `parsed` dict closes that gap: any downstream consumer of the parser result can now read the submission link off the payload directly, which matches the issue's plain-language goal.

## Tests

New regression test in `backend/tests/test_parser_service.py`:

- `test_parse_and_update_stamps_submission_id_onto_parsed_payload` — asserts that after `parse_and_update`, the `parsed` payload received by `update_resume_in_sheet` contains `submission_id` matching the canonical ID passed in.

Existing test `test_parse_and_update_passes_submission_id_and_parsed_to_writeback` (from #154) continues to cover the arg-level threading. `test_parse_and_update_swallows_exceptions` still passes.

Full backend suite: 158/158 green.

## Non-goals (kept strictly out of scope per Kevin's guardrail)

- Logging cleanup / `print` → `logger.info` migration — covered in #130.
- Deterministic Drive filename refactor — #122.
- Append-only submissions write logic — #125.
- Any change to `parse_resume` (the pure parser) or to `update_resume_in_sheet`'s schema/columns.

### DIFF
--- a/backend/services/parser_service.py
+++ b/backend/services/parser_service.py
@@ -5,11 +5,12 @@ from parser import parse_resume
 from storage.sheets_repo import update_resume_in_sheet
 
 
-def parse_and_update(submission_id: str, drive_file_id: str, pre_extracted_text: str = "") -> None:
+def parse_and_update(submission_id: str, drive_file_id: str, pre_extracted_text: str) -> None:
     """Parse the extracted resume text and update the parser_results row in Sheets."""
     try:
         print(f"[JOB] Parsing submission_id={submission_id} drive_file_id={drive_file_id} ...")
         parsed = parse_resume(pre_extracted_text or "")
+        parsed["submission_id"] = submission_id
         parsed["drive_file_id"] = drive_file_id
         update_resume_in_sheet(submission_id, parsed)
         print(f"[JOB] Parsed + updated sheet submission_id={submission_id}")
--- a/backend/tests/test_parser_service.py
+++ b/backend/tests/test_parser_service.py
@@ -27,6 +27,29 @@ def test_parse_and_update_passes_submission_id_and_parsed_to_writeback(monkeypat
     assert captured["parsed"]["drive_file_id"] == "drive-xyz"
 
 
+def test_parse_and_update_stamps_submission_id_onto_parsed_payload(monkeypatch):
+    """#126: the parser context must stamp submission_id onto the parsed payload
+    so the parser's output is reliably linked back to the original submission."""
+    captured = {}
+
+    def fake_parse_resume(text):
+        return {"skills": {"value": ["python"]}}
+
+    def fake_update(submission_id, parsed):
+        captured["parsed"] = parsed
+
+    monkeypatch.setattr(parser_service, "parse_resume", fake_parse_resume)
+    monkeypatch.setattr(parser_service, "update_resume_in_sheet", fake_update)
+
+    parser_service.parse_and_update(
+        submission_id="sub-xyz-canonical",
+        drive_file_id="drive-1",
+        pre_extracted_text="resume text",
+    )
+
+    assert captured["parsed"]["submission_id"] == "sub-xyz-canonical"
+
+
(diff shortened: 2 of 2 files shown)


Kevin-TCUS (approved):
@briannadelariva, this looks good to me. Approved and ready to merge.

This is nicely scoped to #126. The parser job already had submission_id passed through the function chain, and this PR closes the remaining gap by stamping the canonical submission_id directly onto the parsed payload before writeback.

A few things I checked:

* parse_and_update(...) now requires pre_extracted_text explicitly.
* The parsed payload now includes submission_id.
* Existing drive_file_id stamping is preserved.
* update_resume_in_sheet(submission_id, parsed) still receives the canonical ID as the writeback argument.
* The new regression test confirms the parsed payload itself carries the expected submission_id.
* No unrelated parser, schema, logging, Drive naming, or benchmark behavior was changed.

This is a good small step in the v0.3 submission_id lane. Nice work.

PR #255, opened 2026-06-08
Add resolver aware skill scoring in benchmark

### LINKED ISSUE #233: [Benchmark] Add resolver-aware skills scoring alongside raw skills scoring
## Context
The current benchmark may understate parser quality when the parser output and golden JSON use different strings for the same underlying skill.

**Examples:**
* `react.js` vs `React`
* `python3` vs `Python`
* `chash` vs `C#`
* `cplusplus` vs `C++`

Today, raw benchmark scoring may treat those as different skills even when Resolver V1 and the alias map would understand them as equivalent. That can create misleading scoring.

**Example:**
* Gold skill: `React`
* Parser output: `React.js`

Raw benchmark behavior may count this as:
* false positive: `React.js`
* false negative: `React`

But Resolver V1 can map both `React` and `React.js` to the same canonical skill ID: `react`.

This issue is about improving benchmark evaluation fidelity, not changing the parser itself.

## Goal
Add resolver-aware skills scoring to the benchmark alongside the existing raw skills scoring.

The benchmark should help us distinguish between:
1. **Raw extraction quality:** Did the parser extract the literal strings we expected?
2. **Resolved/canonical quality:** Did the parser plus Resolver V1 understand equivalent skill names correctly?

Both views are useful, so this issue should not replace raw scoring.

## Benchmark relevance
This issue addresses a benchmark-observed evaluation problem:
* alias / resolver mismatch
* normalization mismatch
* raw string vs canonical skill comparison

The goal is to avoid optimizing the parser around exact string formatting when the production pipeline may intentionally normalize equivalent terms.

## Existing resolver behavior to remain consistent with
Libelle already has a pure Resolver V1 function: `backend/resolver/resolver.py`

Function: `resolve_extracted_profile(…)`

The resolver:
* takes extracted skills
* normalizes keys using `normalize_key`
* uses the alias map to resolve known aliases
* returns canonical resolved skill IDs
* preserves unknown skills
* reports resolver coverage stats

The alias map currently lives at: `backend/resolver/knowledge/aliases_v1.json`

Examples from the current alias map include:
* `react` -> `react`
* `reactjs` -> `react`
* `react.js` -> `react`
* `python` -> `python`
* `python3` -> `python`
* `cplusplus` -> `c++`
* `chash` -> `c#`

The benchmark should remain consistent with Resolver V1 behavior where practical. Contributors may reuse existing resolver functionality directly or factor shared logic if that produces a cleaner design. 

The benchmark should not introduce a separate alias-resolution system that can drift from Resolver V1 semantics.

## Why not replace raw scoring?
Raw scoring and resolved scoring answer different questions.

**Example:**
* Gold skill: `React`
* Parser output: `React.js`

Resolved scoring may correctly count this as a match because both values resolve to the same canonical skill. However, raw scoring still reveals that the parser emitted a different string than the benchmark expected.

For that reason, resolved scoring should supplement raw scoring, not replace it.

## Examples

**Example 1: React.js vs React**
* **Gold skills:** `React`
* **Parser output:** `React.js`
* **Alias map / resolver behavior:** `React` -> `react`, `React.js` -> `react`
* **Expected benchmark behavior:** Raw scoring may still show a string mismatch. Resolved scoring should count this as a match.

**Example 2: python3 vs Python**
* **Gold skills:** `Python`
* **Parser output:** `python3`
* **Alias map / resolver behavior:** `Python` -> `python`, `python3` -> `python`
* **Expected benchmark behavior:** Raw scoring may show mismatch. Resolved scoring should count this as a match.

**Example 3: chash vs C#**
* **Gold skills:** `C#`
* **Parser output:** `chash`
* **Alias map / resolver behavior:** `C#` -> `c#`, `chash` -> `c#`
* **Expected benchmark behavior:** Raw scoring may show mismatch. Resolved scoring should count this as a match.

**Example 4: unresolved normalization gap**
* **Gold skills:** `PostgreSQL`
* **Parser output:** `postgres`
* **Expected benchmark behavior:** If the current alias map does not map `postgres` to `PostgreSQL`, this should not silently count as a resolved match. Raw scoring may show mismatch. Resolved scoring should also show mismatch unless Resolver V1 / `aliases_v1.json` explicitly supports the mapping. The benchmark should surface this as an alias coverage gap, not hide it.

## Guardrails / Non-goals
**Please keep this issue focused on benchmark scoring.**

This issue should not:
* change parser extraction behavior
* change backend upload behavior
* change live `parser_results` writeback
* change frontend behavior
* rewrite Resolver V1
* add a large skill ontology
* add external APIs
* add AI-based matching
* add fuzzy matching
* silently remove or replace raw benchmark scoring
* modify location scoring
* modify benchmark PDFs
* broadly relabel the golden JSON dataset

The goal is to add a resolved/canonical scoring view alongside the existing raw scoring view.

## Suggested files to look at
**Likely implementation area:**
* `scripts/benchmark.py`

**Existing resolver references:**
* `backend/resolver/resolver.py`
* `backend/resolver/normalize.py`
* `backend/resolver/schemas.py`
* `backend/resolver/knowledge/aliases_v1.json`

**Likely test area:**
* existing benchmark tests, if present
* `backend/tests/` or scripts-related tests, depending on current repo structure (Please inspect the current benchmark structure before deciding where tests belong).

**You probably should not need to modify:**
* `backend/parser.py`
* frontend files
* backend API routes
* Google Sheets code
* benchmark PDF files
* benchmark resume PDFs

Small fixture or alias-map changes are acceptable only if needed to test resolver-aware scoring clearly.

## Acceptance criteria
- [ ] Benchmark continues to report existing raw skills scoring.
- [ ] Benchmark adds a resolver-aware or canonical skills scoring view.
- [ ] Raw scoring and resolved scoring are clearly labeled so they are not confused.
- [ ] Resolver-aware scoring remains consistent with Resolver V1 / `aliases_v1.json` behavior where practical.
- [ ] A resolved score must never count two skills as equivalent unless the current resolver/alias system would also resolve them as equivalent.
- [ ] Equivalent aliases such as `React.js` and `React` can be counted as a resolved match.
- [ ] Equivalent aliases such as `python3` and `Python` can be counted as a resolved match.
- [ ] Unknown or unresolved skills are still reported clearly.
- [ ] Alias gaps remain visible instead of being silently treated as matches.
- [ ] Existing location scoring behavior is not changed.
- [ ] Existing parser behavior is not changed.
- [ ] Tests or fixtures are added for at least 2 alias-equivalence cases.
- [ ] Existing tests pass.
- [ ] Run the benchmark and include a brief before/after note in the PR.

## Suggested output direction
The exact output format can follow the existing benchmark style, but the summary should make the distinction clear.

For example, the benchmark might eventually show something like:
```text
skills_raw Micro-F1: 0.71
skills_resolved Micro-F1: 0.84
```
The exact field names do not have to be these names, but the output should make it clear which score is raw string matching and which score is resolver-aware.

## Suggested test cases

**Alias match: React.js vs React**
* **Gold skills:** `React`
* **Parser output:** `React.js`
* **Expected behavior:** Raw scoring may count mismatch. Resolved scoring should count match if both resolve to the same canonical value.

**Alias match: python3 vs Python**
* **Gold skills:** `Python`
* **Parser output:** `python3`
* **Expected behavior:** Raw scoring may count mismatch. Resolved scoring should count match if both resolve to the same canonical value.

**Alias match: chash vs C#**
* **Gold skills:** `C#`
* **Parser output:** `chash`
* **Expected behavior:** Raw scoring may count mismatch. Resolved scoring should count match if both resolve to the same canonical value.

**Unresolved case**
* **Gold skills:** `PostgreSQL`
* **Parser output:** `database systems`
* **Expected behavior:** Should not be counted as a resolved match unless the resolver/alias data explicitly supports that mapping.

## Notes for implementation
Please keep raw and resolved scoring separate. Raw scoring tells us how closely parser output matches the gold strings. Resolved scoring tells us whether the parser/resolver pipeline understands equivalent terms. Both are useful.

The benchmark should remain consistent with Resolver V1 behavior where practical. Contributors may reuse existing resolver functionality directly or factor shared logic if that produces a cleaner design. The benchmark should not introduce a separate alias-resolution system that can drift from Resolver V1 semantics.

If the current alias map does not contain enough aliases to demonstrate the behavior, keep any alias additions small and test-driven.

## Why this matters
Without resolver-aware scoring, the benchmark may make the parser look worse than it is in cases where the output is semantically correct but formatted differently.

At the same time, raw scoring is still valuable because it shows exactly what the parser extracted.

Keeping both views helps us separate:
1. parser extraction problems
2. normalization problems
3. resolver/alias coverage gaps
4. true missing skills

This will make future parser and resolver work easier to prioritize.

jriya186:
Closes #233 

## Summary

Adds resolver-aware skills scoring (`skills_resolved`) to the benchmark alongside the existing raw skills scoring (`skills`). The two views answer distinct questions:

- **`skills` (raw):** Did the parser extract the exact strings we expected?
- **`skills_resolved`:** Did the parser + Resolver V1 understand equivalent skill names correctly?

## Changes

- Added `_resolve_skill()` and `_resolve_skill_set()` to `scripts/benchmark.py` — thin helpers that run `normalize_key` + alias map lookup, reusing Resolver V1 semantics directly without introducing a separate resolution system
- Added `score_skills_resolved()` — same TP/FP/FN structure as `score_skills()` but operates on resolved canonical IDs
- Added a `skills_resolved` scoring block in the main loop, appended after the existing raw skills block. Raw field name (`skills`) is unchanged to avoid breaking anything downstream.
- Added `Set` to typing imports
- Added `backend/tests/test_benchmark_resolved_scoring.py` with 3 tests: `React.js` → `React` alias match, `python3` → `Python` alias match, and an unresolved case (`database systems` vs `PostgreSQL`) that must not count as a match

## Benchmark

**Hand-built corpus (10 resumes)**

| Parser | Field | Micro-F1 | Macro-F1 |
|--------|-------|----------|----------|
| libelle | skills | 0.768 | 0.767 |
| libelle | skills_resolved | 0.768 | 0.767 |
| libelle | location | 0.615 | 0.400 |

Scores are identical on the hand-built corpus — the existing mismatches are parser extraction failures and delimiter bugs, not alias-equivalence mismatches.

**Synthetic corpus (30 resumes, --seed 42 --count 30 --adversarial-ratio 0.34)**

| Parser | Field | Micro-F1 | Macro-F1 |
|--------|-------|----------|----------|
| libelle | skills | 0.656 | 0.565 |
| libelle | skills_resolved | 0.694 | 0.594 |
| libelle | location | 0.889 | 0.800 |

`skills_resolved` scores higher than `skills` on the synthetic corpus (0.694 vs 0.656 Micro-F1), confirming that the resolved scoring is correctly giving credit for alias-equivalent skill names that the raw scoring would penalize.

## Tests

```
cd backend && pytest tests/
166 passed in 0.53s
```

## What was not changed

- Parser extraction behavior
- Resolver V1 implementation
- Location scoring
- Golden JSON dataset
- Alias map — `react.js`, `python3`, and `chash` were already present in `aliases_v1.json`
- `report.csv` column schema for existing fields

### DIFF
--- a/backend/tests/test_benchmark_resolved_scoring.py
+++ b/backend/tests/test_benchmark_resolved_scoring.py
@@ -1,0 +1,43 @@
+import sys
+from pathlib import Path
+
+BACKEND_DIR = Path(__file__).parent.parent
+if str(BACKEND_DIR) not in sys.path:
+    sys.path.insert(0, str(BACKEND_DIR))
+
+SCRIPTS_DIR = Path(__file__).parent.parent.parent / "scripts"
+if str(SCRIPTS_DIR) not in sys.path:
+    sys.path.insert(0, str(SCRIPTS_DIR))
+
+from resolver.normalize import normalize_key
+from benchmark import score_skills_resolved
+
+ALIASES = {
+    normalize_key("react"): "react",
+    normalize_key("react.js"): "react",
+    normalize_key("reactjs"): "react",
+    normalize_key("python"): "python",
+    normalize_key("python3"): "python",
+    normalize_key("chash"): "c#",
+    normalize_key("c#"): "c#",
+}
+
+
+def test_reactjs_resolves_to_react():
+    result = score_skills_resolved(["React.js"], ["React"], ALIASES)
+    assert result["tp_count"] == 1
+    assert result["fp_count"] == 0
+    assert result["fn_count"] == 0
+
+
+def test_python3_resolves_to_python():
+    result = score_skills_resolved(["python3"], ["Python"], ALIASES)
+    assert result["tp_count"] == 1
+    assert result["fp_count"] == 0
+    assert result["fn_count"] == 0
+
+
+def test_unresolved_skills_do_not_match():
+    result = score_skills_resolved(["database systems"], ["PostgreSQL"], ALIASES)
+    assert result["tp_count"] == 0
+    assert result["fn_count"] == 1
--- a/scripts/benchmark.py
+++ b/scripts/benchmark.py
@@ -24,7 +24,7 @@ import traceback
 from collections import Counter, defaultdict
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Tuple, Set
 
 
 
@@ -309,6 +309,21 @@ def _normalize(s: Any) -> str:
         return ""
     return str(s).strip().lower()
 
+def _resolve_skill(skill: str, aliases: Dict[str, str]) -> str:
+    """Resolve a single skill to its canonical ID via the alias map.
+    Falls through to normalized key if no alias exists."""
+    from resolver.normalize import normalize_key
+    key = normalize_key(skill)
+    return aliases.get(key, key) if key else ""
+
+def _resolve_skill_set(skills: List[str], aliases: Dict[str, str]) -> Set[str]:
+    resolved = set()
+    for s in skills:
+        if s:
+            r = _resolve_skill(s, aliases)
+            if r:
+                resolved.add(r)
+    return resolved
 
 def score_skills(
     predicted: List[str], golden: List[str]
@@ -329,6 +344,26 @@ def score_skills(
         "fn_examples": sorted(fn)[:10],
     }
 
+def score_skills_resolved(
+    predicted: List[str],
+    golden: List[str],
+    aliases: Dict[str, str],
+) -> Dict[str, Any]:
+    pred_set = _resolve_skill_set(predicted, aliases)
+    gold_set = _resolve_skill_set(golden, aliases)
+
+    tp = pred_set & gold_set
+    fp = pred_set - gold_set
+    fn = gold_set - pred_set
+
+    return {
+        "tp_count": len(tp),
+        "fp_count": len(fp),
+        "fn_count": len(fn),
+        "tp_examples": sorted(tp)[:10],
+        "fp_examples": sorted(fp)[:10],
+        "fn_examples": sorted(fn)[:10],
+    }
 
 def score_location(
     predicted: Dict[str, str], golden: Dict[str, str]
@@ -725,7 +760,7 @@ def main() -> None:
 
             print(f"done ({runtime_ms:.0f}ms)")
 
-            # Score skills
+            # Score skills (raw)
             skills_score = score_skills(
                 adapted["skills"], golden.get("skills", [])
             )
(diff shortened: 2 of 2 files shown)


Kevin-TCUS (approved):
@jriya186, this looks excellent. Approved and ready to merge.

You executed the scope of #233 very well. Keeping both skills raw scoring and skills_resolved resolver-aware scoring side by side gives us exactly the benchmark visibility we needed.

A few specific things that make this a strong PR:

1. Raw skills scoring remains unchanged, so we do not lose historical comparability.
2. skills_resolved gives us a separate canonical/alias-aware view without hiding raw extraction mismatches.
3. _resolve_skill() falls through to the normalized key when no alias exists, which is the right behavior. Exact normalized matches can still count, while different unresolved terms still fail.
4. The TP/FP/FN set logic is clean and easy to follow.
5. The benchmark results make sense: the hand-built corpus staying the same suggests those misses are mostly extraction/delimiter issues, while the synthetic corpus improvement shows resolved scoring is giving credit for alias-equivalent skill names.

Two small future cleanup notes, not blockers:

- In test_unresolved_skills_do_not_match, we could also assert fp_count == 1 to make the test fully explicit.
- The alias path lookup in benchmark.py is broad. That is fine for CLI flexibility now, but later we may want to centralize alias loading with runtime resolver helpers to avoid drift.

Neither is blocking. This is a solid evaluation improvement.

PR #274, opened 2026-06-18
PyMuPDF x/y multi-column layout spike findings

jriya186:
## Summary

Research spike to determine whether PyMuPDF x/y coordinate and block-level metadata can explain benchmark reading-order failures on multi-column resumes, or whether current issues are still solvable with text-level parser heuristics.

This is observation only. No production code was modified.

## What's included

- `backend/benchmarks/layout_spike/PYMUPDF_XY_MULTICOLUMN_FINDINGS.md` — full findings doc following the required structure (files inspected, extraction modes, plain text vs x/y comparison, multi-column findings, secondary observations, text-only sufficiency check, recommendation)
- `backend/benchmarks/layout_spike/inspect_pymupdf_layout.py` — exploratory script covering all extraction modes tested (plain text, current production blocks, column-aware blocks, dict view with header detection, table detection, header-scoped grouping)
- `backend/benchmarks/layout_spike/test_layout_extraction.py` — lighter CLI-runnable script covering just the two modes that proved useful (column-aware, header-scoped), for quick ad hoc testing

Both scripts take a PDF path as a command-line argument and are not imported by any production code path.

## Files inspected

- Two multi-column resumes (one floating text-box layout, one sidebar layout) — suspected reading-order issues
- One clean single-column resume (control)
- One dense-skills resume
- One project-heavy/academic resume

(One of the multi-column files is a real volunteer resume, not synthetic data — referenced in the findings doc as "volunteer resume" with no identifying filename or details, and not included in this PR.)

## Key findings

- Confirmed: current production extraction (`get_text("blocks")` sorted globally by `(y0, x0)`) interleaves columns on both multi-column resumes tested — e.g. an Education entry appearing mid-block inside an Experience entry.
- Column-aware sorting (x0 clustering into lanes, sorted independently) and header-scoped grouping (bucketing content under its nearest header per lane) both resolved this interleaving cleanly on the multi-column files.
- However, the column-boundary heuristic produced false positives on all 3 single-column resumes tested — usually cosmetic (an isolated name or contact block misclassified as a second column), but in one case it broke header-to-content association entirely.
- `page.find_tables()` is not available in our installed PyMuPDF version — noted but not evaluated.

## Recommendation

**Useful later, but not needed for v0.3.**

The x/y approach clearly fixes a real, confirmed extraction-layer information-loss problem on multi-column resumes. But it's not reliable enough to apply unconditionally yet — the false-positive rate on single-column resumes (the majority of our corpus) would need to be addressed first (e.g. minimum-blocks-per-lane checks, vertical-overlap validation between lanes) before this could be safely introduced to production.

Full reasoning and per-file breakdown in the findings doc.

## Guardrails followed

- No changes to `parser.py`, `scripts/benchmark.py`, resolver logic, or benchmark PDFs/golden JSON
- No new runtime dependencies (fitz only)
- No layout-aware parser, no `Document` abstraction introduced
- PyMuPDF's `multi_column.py` utility was used only as a local reference for conceptual comparison — not copied into the repo, per the licensing guardrail in the issue

## Acceptance criteria

- [x] Findings doc created at the specified path, following the required structure
- [x] Exploratory script(s) included, clearly marked exploratory, not wired into production
- [x] At least one PDF with confirmed plain-text extraction issues included in the inspection
- [x] Recommendation tied to inspected files, not general PDF theory
- [x] No production code changes

### DIFF
--- a/backend/benchmarks/layout_spike/PYMUPDF_XY_MULTICOLUMN_FINDINGS.md
+++ b/backend/benchmarks/layout_spike/PYMUPDF_XY_MULTICOLUMN_FINDINGS.md
@@ -1,0 +1,116 @@
+# PyMuPDF X/Y & Multi-Column Layout Spike — Findings
+
+**Status:** Observation only. No production code was modified as part of this spike.
+**Related issue:** PyMuPDF x/y & multi-column layout research spike
+
+---
+
+## 1. Files inspected
+
+| File | Why selected |
+|---|---|
+| `multi_col_02.pdf` (Lina Calder) | Suspected multi-column / reading-order issue — floating text-box layout |
+| Volunteer resume (real, non-synthetic — filename withheld) | Suspected multi-column / reading-order issue — sidebar-style two-column layout |
+| `embed_link_01.pdf` (Daniel Foster) | Clean single-column control |
+| `high_signal_01.pdf` (Rohan Mehta) | Dense skills resume, with sub-header-labeled skill groups |
+| `Jadhav_Riya.pdf` | Project-heavy / academic-style resume |
+
+---
+
+## 2. PyMuPDF extraction modes inspected
+
+- `get_text("text")` — plain text, raw storage order
+- `get_text("blocks")` — current production extraction (sorted globally by `(y0, x0)`)
+- `get_text("blocks")` + x0 clustering — custom column-aware sort (per-lane `y0` sort)
+- `get_text("dict")` — block/line/span structure with bounding boxes
+- `page.find_tables()` — attempted; **not available** in the installed PyMuPDF version (`AttributeError: 'Page' object has no attribute 'find_tables'`). Requires PyMuPDF 1.23+. Not evaluated further in this spike.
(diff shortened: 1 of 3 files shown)


Kevin-TCUS (approved):
@jriya186, this is excellent work. Approved.

This is exactly the kind of research spike we needed for the multi-column resume question: bounded, evidence-driven, isolated from production code, and honest about the tradeoffs.

A few things that make this especially strong:

1. Good engineering judgment
    You confirmed that x/y coordinate data can fix real multi-column reading-order failures, but you still recommended against shipping it for v0.3 because the heuristic regressed single-column controls. That restraint is important.
2. Clean isolation
    No production parser, resolver, intake, benchmark, or dashboard behavior was changed. Keeping this under backend/benchmarks/layout_spike avoids destabilizing the current v0.3 pipeline.
3. Useful reproducibility
    Including both the fuller inspection script and the lighter CLI tester makes the findings easier to revisit later without turning them into production behavior now.
4. Clear recommendation
    “Useful later, but not needed for v0.3” is the right conclusion. This gives us a future path for layout-aware extraction while keeping the current parser work focused.

This gives us the diagnostic clarity we needed without creating scope creep. Great work.

PR #318, opened 2026-07-10
Add multi-column synthetic corpus + layout-aware extraction prototype (#275)

### LINKED ISSUE #275: [Parser Spike] Harden layout-aware extraction for multi-column resumes
## Context
PR #274 added an observation-only PyMuPDF x/y layout spike under `backend/benchmarks/layout_spike/`.

The spike found that x/y coordinate data can help with genuine multi-column resume reading-order failures, but the current exploratory column-detection heuristic is not safe enough for production.

**Key findings:**
* Layout-aware extraction helped true multi-column resumes.
* The exploratory x/y column-boundary heuristic false-positived on single-column resumes.
* In one control case, the false positive damaged header-to-content association.
* Therefore, this should not be introduced into the v0.3 parser pipeline yet.

This issue preserves the future work without expanding v0.3 scope.

## Goal
Design and eventually prototype a safer layout-aware extraction path for multi-column resumes, without regressing normal single-column resumes. 

The goal is not to replace the current parser immediately. The goal is to determine what safeguards would be required before layout-aware extraction could be considered safe.

## Why This Matters
Some multi-column resumes lose important structure when text is extracted in global reading order. Once columns are interleaved, downstream parser heuristics may not be able to recover the intended human reading order. 

However, most benchmark resumes appear to be single-column or sufficiently handled by current block sorting and text-level heuristics. A layout-aware path must therefore avoid harming the majority case.

## Required Future Safeguards
A future prototype should evaluate safeguards such as:
* Minimum block count per detected lane.
* Vertical-overlap validation between candidate columns.
* Ignoring isolated name/contact/header blocks when deciding columns.
* Requiring sustained two-column structure across meaningful page regions.
* Fallback to current production extraction when confidence is low.
* Comparison against current production extraction before accepting layout-aware output.

## Suggested Future Work
1. Build a small layout-aware extraction prototype behind an explicit experimental flag.
2. Test it on known multi-column benchmark resumes, single-column controls, dense skills resumes, project-heavy resumes, and contact-heavy resumes.
3. Measure whether it improves multi-column cases without regressing single-column cases.
4. Produce a recommendation: adopt, keep experimental, or defer again.

## Non-Goals
* No v0.3 parser rewrite.
* No unconditional replacement of current extraction.
* No new production dependency without review.
* No layout-aware extraction enabled by default.
* No changes to intake, dashboard, resolver, or Sheets behavior.
* No use of private resume content in committed artifacts.
* No claim of improvement without benchmark evidence.

## Acceptance Criteria
- [x] Uses PR #274 findings as the starting evidence.
- [x] Evaluates true multi-column resumes and single-column controls.
- [x] Defines safeguards for avoiding false-positive column detection.
- [ ] Compares layout-aware extraction against current production extraction.
- [x] Documents regressions as well as improvements.
- [x] Recommends adopt, keep experimental, or defer.
- [x] Does not enable layout-aware extraction in production by default.

jriya186:
## Summary

Closes #275.

Expands the resume parsing benchmark corpus with 10 new synthetic resumes and 1 replaced resume, adds a hardened layout-aware extraction prototype with measured safeguards, and produces a benchmark-evidence-backed recommendation report comparing it against current production extraction.

This PR does **not** change any production code path. `scripts/benchmark.py`, `parser.py`, `intake_service.py`, and the resolver are all untouched.

## What's in this PR

**Corpus additions** (`backend/benchmarks/resumes/` + `backend/benchmarks/golden_json/`)

- 10 new synthetic resumes across true multi-column, single-column layout-trap, sparse-skills, dense-skills, project-heavy, and contact-heavy categories
- `multi_col_01.pdf` replaced — old version had no selectable text layer (image-only); new version is text-based. Existing V1 golden JSON for this file is left untouched, since it was validated against the original PR #274 spike run.
- All new golden JSONs follow the V2 `sections` format (per Kevin's request), with `title`/`meta`/`subtitle`/`bullets` structure for experience/education/project entries

**Layout-aware extraction prototype** (`backend/benchmarks/layout_aware_extraction/`)

- `layout_aware_extraction.py` — hardened column-detection logic building on the PR #274 spike's `_detect_column_boundary` / header-scoped grouping, gated behind an explicit `enabled` flag (default `False`, never called from production)
- Implements 4 safeguards absent from the #274 spike code: minimum block count per lane, vertical-overlap validation, isolated header/contact-block exclusion, sustained two-column structure requirement, plus fallback to production ordering when any safeguard fails
- `compare_layout_aware.py` — standalone comparison script. Imports `parse_resume()` and the scoring functions from `scripts/benchmark.py` unmodified; runs the full 20-resume corpus through both current production extraction and the layout-aware path, scores both against the same golden JSONs, and reports per-resume F1 deltas

**Recommendation report** (`backend/benchmarks/layout_spike/LAYOUT_AWARE_EXTRACTION_RECOMMENDATION.md`)

## Benchmark results

Full corpus (20 resumes: 10 true multi-column, 10 true single-column) run through `compare_layout_aware.py`:

| Metric | Result |
|---|---|
| False positives on single-column resumes | **0 / 10** |
| True multi-column resumes triggering layout-aware extraction | **5 / 10** |
| Of those 5: improvements | **2** (`header_contact_01` +0.517, `multi_col_01` +0.167) |
| Of those 5: regressions | **2** (`multi_col_02` -0.258, `sparse_skill_01` -0.400) |
| Of those 5: not meaningful | **1** (`multi_col_04` — golden JSON has empty skills array regardless of extraction quality) |

Both regressions show the same failure signature: column detection correctly identifies the multi-column structure, but downstream skill extraction still drops to F1 = 0.0. Root cause is very likely the header-scoped text reconstruction stripping commas/delimiters when flattening multi-item skill lines to one item per line — a text-formatting bug, not a detection failure. Full breakdown, per-resume table, and root-cause discussion in the report.

## Recommendation

**Keep experimental — do not adopt.**

Safeguards are safe on single-column resumes (zero false positives) but too conservative on true multi-column resumes (miss half of them), and even when correctly triggered, results are inconsistent. Next steps before revisiting adoption: fix the header-scoped text-assembly bug, retune safeguard thresholds against the 5 missed multi-column cases, and re-run the comparison.

## Testing

- `compare_layout_aware.py` run against the full corpus (see results above and full CSV/summary in report)
- All new synthetic resumes are grounded directly in their PDF content — no filename-to-person assumptions
- No private or real resume content committed

## Non-goals confirmed

- [x] No v0.3 parser rewrite
- [x] No unconditional replacement of current extraction
- [x] No new production dependency
- [x] Layout-aware extraction not enabled by default anywhere
- [x] No changes to intake, dashboard, resolver, or Sheets behavior
- [x] No private resume content in committed artifacts
- [x] No claim of improvement without benchmark evidence


Kevin-TCUS (changes requested):
## PR Review — Part 1: Synthetic Resume Corpus and Golden JSONs

Hi Riya,

This is a strong expansion of the benchmark corpus. The new resumes cover useful failure modes across true multi-column layouts, dense and sparse skills, project-heavy content, contact-heavy headers, and single-column layout traps.

A quick note on why we are being careful here: these golden JSON files become the benchmark’s expected truth. If a golden is mislabeled or inconsistent, a correct parser can look wrong later. The goal is not only to fix this batch, but also to make the labeling rules clearer and easier to repeat for future contributors.

Before we merge, could you please work through this checklist?

### Required fixes

- [x] **`multi_col_04`**
  - Change `resume_id` from `multi_col_05` to `multi_col_04`
  - Add the professional summary from the top of the PDF to the `sections` array

- [x] **`multi_col_05`**
  - Change `resume_id` from `multi_col_06` to `multi_col_05`

### Skills cleanup

- [x] Remove standalone modifiers such as:
  - `"basic"` from `JavaScript (basic)`
  - `"advanced"` from `Excel (advanced, pivot tables, VBA)`

Why: `basic` and `advanced` describe proficiency level; they are not skills on their own.

Terms such as `Pivot Tables`, `VBA`, `EPA`, `NEPA`, `GWAS`, and `ACE/ADE` can remain as standalone normalized values because they are meaningful tools, methods, acronyms, or domain concepts.

For future consistency, it would also be useful to confirm whether the normalized `skills` array should:

1. preserve relationships such as `Ad Copy (Google, Meta, LinkedIn)`, or
2. flatten meaningful tools, platforms, acronyms, and methods into separate values.

Either approach can work, but the rule should be applied consistently.

### `dense_skills_01`

- [x] Confirm whether `Information` is intentionally truncated in the PDF
  - If intentional, keep it and note that clearly
  - If not, update both the PDF and JSON to `Information Architecture`

Why: the golden should represent either a deliberate benchmark trap or a corrected source, rather than an accidental ambiguity.

- [x] Confirm whether education locations are intentionally excluded from V2
  - The PDF includes `Santa Barbara, CA`, but the JSON does not

Why: either choice may be valid, but the rule should be explicit so future goldens handle education locations consistently.

### Files that looked good

These looked accurate and approval-ready:

- `multi_col_01` — intentional legacy V1 fixture from the earlier #274 spike
- `dense_skills_02`
- `header_contact_01`
- `multi_col_03`
- `project_heavy_01`
- `sparse_skill_01`

The two single-column layout-trap files also look structurally good once the standalone `"advanced"` modifier is removed.

## Review Summary

| Fixture | PDF | JSON | Status | Notes |
|---|---:|---:|---|---|
| `multi_col_01` | Present | Present | **Pass** | Intentional legacy V1 fixture from the #274 spike |
| `dense_skills_01` | Present | Present | **Change** | Remove `"basic"` as a standalone skill; clarify `Information` and education-location handling |
| `dense_skills_02` | Present | Present | **Pass** | Good V2 mapping across summary, skills, experience, and education |
| `header_contact_01` | Present | Present | **Pass** | Strong source-to-golden alignment |
| `multi_col_03` | Present | Present | **Pass** | Accurate two-column reconstruction |
| `multi_col_04` | Present | Present | **Change required** | Wrong `resume_id`; professional summary missing from `sections` |
| `multi_col_05` | Present | Present | **Change required** | Wrong `resume_id` |
| `project_heavy_01` | Present | Present | **Pass** | Accurate project-heavy mapping |
| `single_col_layout_trap_01` | Present | Present | **Change** | Remove `"advanced"` as a standalone skill |
| `single_col_layout_trap_02` | Present | Present | **Change** | Remove `"advanced"` as a standalone skill |
| `sparse_skill_01` | Present | Present | **Pass** | Complete V2 golden; intentionally preserves project-like content under the source `CERTIFICATIONS` heading |

Once the required fixes and normalization clarifications are complete, this portion should be ready to approve.
## PR Review — Part 2: Layout-Aware Extraction Prototype and Benchmark Evidence

Hi Riya,

I reviewed the layout-aware extraction prototype, comparison script, and recommendation report together. I also used the current `scripts/benchmark.py` only as reference context to understand what the comparison is reusing; I am not requesting changes to that existing benchmark file.

Overall, this is a valuable spike. The decision to keep the prototype experimental is supported by the current results, and the comparison harness makes a good effort to keep production and layout-aware scoring apples-to-apples.

Before we treat the report as final, I think a few code and evidence issues should be addressed.

### Required Code Fixes

- [x] **Fix uppercase header detection in `_is_header()`**

    The current logic lowercases the text before checking `.isupper()`:
    ```python
    cleaned = text.strip().rstrip(":").lower()

    if cleaned.replace(" ", "").replace("&", "").isupper():
        return True 
    ```
    After `.lower()`, `.isupper()` can never succeed. 
    **Why this matters:** Several legitimate corpus headers may not be covered by the exact keyword list and may therefore depend on this uppercase-detection path. That can make true multi-column resumes fail the “headers in both lanes” safeguard. Please check uppercase status before lowercasing, then use a normalized lowercase value for keyword matching.

- [x] **Review header coverage before tuning thresholds**

    Please confirm recognition for headings used in this corpus, including examples such as: `Profile`, `Career Summary`, `Key Skills`, `Areas of Expertise`, and `Short Courses`.
    **Why:** Some missed multi-column detections may be caused by header classification, not only by thresholds such as minimum blocks or vertical overlap. It would be better to fix header recognition first and rerun before retuning safeguards.

- [x] **Clarify or simplify text reconstruction**

    Rename or implement `_header_scoped_text()` more fully. The current function appears to order all left-lane blocks by vertical position, followed by all right-lane blocks. It assigns `current_header`, but that value is not used to group subsequent content. 
    **Why:** The name and comments currently imply content is grouped under the nearest header within each lane, while the implementation is closer to lane-aware reading-order reconstruction. Either approach is reasonable for a prototype. The code and documentation should simply describe the same behavior.
    *Tip: Remove the unused `current_header` variable if true grouping is not intended. This will make the prototype easier for the next contributor to understand.*

### Strengthen the Regression Evidence

- [ ] **Reword delimiter loss as a hypothesis until diagnostics confirm it**

    The report currently identifies delimiter loss in the reconstructed text as the likely root cause of the `multi_col_02` and `sparse_skill_01` regressions. That is a reasonable hypothesis, but the current code does not explicitly strip commas or delimiters, so I do not think the root cause is confirmed yet. 
    **Suggested wording:** *"The shared failure signature suggests that reconstructed text formatting or ordering is breaking downstream skill extraction. Delimiter loss is one possible cause and should be confirmed through side-by-side diagnostic output."*

- [ ] **Add diagnostic output for triggered cases**

    For each resume where layout-aware extraction is used, it would help to record the production input text, layout-aware reconstructed text, production extracted skills, layout-aware extracted skills, false-positive and false-negative skills, detected boundary, and safeguard values with pass/fail reasons.
    **Why:** This will let the report distinguish among detection failure, reading-order failure, punctuation or line-break changes, and downstream parser sensitivity.

### Benchmark Integrity

- [x] **Fail or clearly surface missing golden files in `compare_layout_aware.py`**

    The comparison script currently warns and skips when a golden JSON is missing. For an evidence-generating benchmark, a run should not quietly finish while omitting part of the claimed corpus.
    **Suggested behavior:** Fail by default when a PDF or golden is missing, OR require an explicit `--allow-missing-goldens` option and list skipped fixtures prominently in the summary.
    **Why:** The report should only claim a 20-resume run when all 20 pairs were actually scored.

- [ ] **Add a guard for an empty result set**

    The CSV-writing path (`writer = csv.DictWriter(f, fieldnames=list(rows[0].keys()))`) assumes at least one result exists. A clear error would be safer than indexing the first row when nothing was scored.

- [ ] **Clarify what the comparison measures**

    State explicitly that this comparison scores skills and location only. The comparison reuses the existing benchmark scoring functions, which currently evaluate only top-level skills and location. 
    **Suggested report note:** *"This experiment measures skills and location extraction only. The V2 sections annotations are not scored in the current comparison and are retained for future benchmark expansion."*

### Recommendation Wording

I agree with: *"Keep experimental — do not adopt."* The current evidence supports that conclusion.

However, I would soften the statement that the “core column-detection approach is sound.” The current corpus shows zero false positives on 10 single-column controls, but layout-aware triggering on only 5 of 10 true multi-column cases, two improvements, two regressions, and one case that is not measurable.
    
**Suggested wording:** *"The approach shows promise, and the safeguards appear conservative on this corpus, but both detection coverage and text reconstruction require further validation before adoption."*

### Suggested Order of Work

1.  Fix `_is_header()` and confirm corpus header coverage.
2.  Rerun the full comparison without changing thresholds.
3.  Add side-by-side diagnostics for triggered resumes.
4.  Confirm the actual cause of the two regressions.
5.  Fix reconstruction behavior if needed.
6.  Retune safeguards *only after* the header and reconstruction issues are understood.
7.  Update the recommendation report with the new evidence.

### Review Summary

| Area | Status | Comment |
| :--- | :--- | :--- |
| **Experimental isolation** | ✅ Pass | Prototype is gated and does not alter the production path. |
| **Production comparison fallback** | ✅ Pass | Fallback uses the exact production text, supporting zero-delta control results. |
| **Header detection** | ❌ Change required | Uppercase detection is currently unreachable after lowercasing. |
| **Header coverage** | ❌ Change required | Confirm corpus headings before concluding thresholds are the main cause of missed detections. |
| **Lane reconstruction** | ⚠️ Clarify/Change | Implementation does not currently perform the header-scoped grouping described by its name and comments. |
| **Root-cause evidence** | ❌ Change required | Formatting/delimiter explanation is still a hypothesis without side-by-side diagnostics. |
| **Missing golden handling** | ❌ Change required | Evidence runs should fail or prominently report incomplete corpus coverage. |
| **Benchmark scope** | ⚠️ Clarify | Existing scoring covers skills and location, not the V2 sections structure. |

**Overall:** Request changes. The spike is useful, and the experimental recommendation is directionally right. The header-detection bug and evidence gaps should be resolved before the report is treated as final.

jriya186:
Hi Kevin,

Pushed fixes for everything flagged. Summary:

## Golden JSONs

- `multi_col_04`: fixed `resume_id`, added the missing professional summary to `sections`
- `multi_col_05`: fixed `resume_id`
- `single_col_layout_trap_01`/`02`: removed `"advanced"` as a standalone skill (kept `Excel`, `pivot tables`, `VBA` separate)
- `dense_skills_01`: removed `"basic"` as a standalone skill; fixed the source PDF so "Information Architecture" is no longer truncated (removes the ambiguity rather than just documenting it); added an explicit note clarifying that the Santa Barbara education location is intentionally omitted per the existing v1 location rule (header/contact preferred, secondary locations dropped) — v2 doesn't redefine this, so v1 still applies

**Open item, not fixed:** your question on whether the skills array should preserve parenthetical groupings (`Ad Copy (Google, Meta, LinkedIn)`) or flatten them — I haven't standardized this across the corpus yet. Didn't want to guess at a rule change without your input; happy to align on one approach and do a pass if you want.

## Layout-aware extraction prototype

- `_is_header()`: fixed — it was lowercasing text before checking `.isupper()`, so the uppercase path was dead code. Also expanded `HEADER_KEYWORDS` to cover `profile`, `career summary`, `key skills`, `areas of expertise`, `short courses`, plus the v2-slice equivalents (`technical toolkit`, `core competencies`, `specialised skills`)
- Added diagnostics (`diagnostics.json`) for every resume where layout-aware extraction triggers — production text, LA text, extracted skills, and exact skill-level diffs against gold
- Used those diagnostics to actually trace both regressions rather than guess: both `multi_col_02` and `sparse_skill_01` were failing because `_get_flat_blocks()` flattened all lines within a PyMuPDF block into one space-joined string, destroying the line breaks between individual skill items (and, for `sparse_skill_01`, between a trailing content line and the following header). Fixed by preserving line breaks in that function.
- Result: both are now improvements, not regressions — `multi_col_02` went from -0.258 to **+0.123**, `sparse_skill_01` from -0.400 to **+0.156**, both with zero skills lost (`lost_in_la: []` in diagnostics). Zero new false positives introduced anywhere in the single-column corpus (still 0/10).
- Also rewrote `_header_scoped_text()` so it actually groups content under headers per lane, matching what its docstring claimed
- `compare_layout_aware.py`: now fails loudly on missing golden JSONs unless `--allow-missing-goldens` is passed, and guards against an empty result set before writing the CSV

## Report

- Updated with the confirmed (not hypothesized) root cause, the current full results table, the skills/location-only scope note, and softened the recommendation to "shows promise, but detection coverage and reconstruction needed further validation" per your wording

**Still open, not yet diagnosed:** 4 of the 10 true multi-column resumes (`dense_skills_01`, `dense_skills_02`, `multi_col_05`, `project_heavy_01`) still fall back to production instead of triggering layout-aware extraction. Per your suggested order, I haven't touched thresholds since I haven't root-caused this the same way as the other two — flagging it as the next thing to look at rather than guessing at a fix.

Given the scope of this one, please flag anything else you think is worth a closer look. I'm happy to take another pass on anything I may have missed.

Kevin-TCUS (changes requested):
Hi Riya,

Thank you for the thorough revision. I reviewed the updated corpus, layout-aware prototype, comparison harness, and recommendation report together.

The technical response to the earlier review is excellent. You fixed the uppercase-header bug, expanded header coverage, added useful per-fixture diagnostics, and traced the two regressions to the actual block-flattening defect rather than leaving the cause as a hypothesis. Preserving internal line breaks in `_get_flat_blocks()` converted both regressions into improvements, and the stronger missing-golden and empty-run behavior makes the comparison much more trustworthy. Keeping the prototype experimental is also the right recommendation.

You also intentionally left the remaining annotation-policy question open rather than guessing. Thank you for doing that. PR #345 has now finalized the canonical V2 rules, so the guidance below resolves that question and reconciles the older fixtures with the new contract.

Before merging, I have a few focused requests.

### 1. Resolve the grouped-skills normalization question

You asked whether parenthetical groupings should be preserved or flattened.

The canonical V2 rule is to preserve the source-faithful grouped wording in `sections`, while extracting the parent concept and independently meaningful parenthetical items into the normalized root `skills` array.

You already applied this correctly in `dense_skills_02.json`: `Ad Copy (Google, Meta, LinkedIn)` is preserved in `sections`, while the root array contains `Ad Copy`, `Google`, `Meta`, and `LinkedIn`.

Please use that as the standard and apply it consistently to any remaining affected fixtures.

### 2. Align the V2 goldens with the canonical specification

Because this PR began before the V2 specification was finalized, two fixtures still carry V1-era assumptions.

**`dense_skills_01`:** Keep San Francisco as the root candidate location, but preserve the source's `Santa Barbara, CA` education location inside the structured education entry. The V2 contract distinguishes the root candidate location from secondary education locations. Secondary locations should not become the root location, but they should remain in the source-faithful structured section.

**`multi_col_04`:** The root `skills` array is currently empty because the heading is `AREAS OF EXPERTISE`. Under the canonical V2 specification, equivalent headings count when the content is primarily tools, platforms, methods, or professional competencies.

This section contains explicit benchmarkable concepts including HubSpot, Mailchimp, Google Analytics, Adobe Creative Suite, Brand Strategy, and SEO/SEM. Please populate the root `skills` array using the V2 normalization rules.

Because this changes the scored golden truth, please rerun the comparison and update the report's current "not measurable" conclusion.

### 3. Correct the trigger-count inconsistency

The report says that 6 of 10 true multi-column resumes trigger layout-aware extraction, which means four fall back. However, the following sentence lists five fallback fixtures and includes `multi_col_04`, even though `multi_col_04` appears in the triggered-results table.

Please remove `multi_col_04` from the fallback list and make the trigger count, fallback list, results table, and recommendation text internally consistent.

### 4. Update the `_header_scoped_text()` documentation

After the `_get_flat_blocks()` fix, block-internal line breaks are intentionally preserved. However, the `_header_scoped_text()` docstring still says each block is preserved as "a single line."

Please update that wording so it accurately reflects the implementation—for example, that each block's internal line structure is preserved and only block ordering and grouping change.

### 5. Document the mixed-page fallback limitation

The comparison harness guarantees exact production text when the entire document falls back. For a mixed multi-page document, however, the document-level `any(...)` flag causes the joined prototype output to be used even for pages where layout-aware extraction was rejected.

If all fixtures in the current comparison are single-page, this does not invalidate the reported run and does not require a per-page implementation change in this PR. Please narrow the "exact production fallback" wording to the behavior actually guaranteed here and record mixed-page fallback as follow-up work.

After the golden updates, please rerun the full comparison and update the results table, counts, diagnostics-derived conclusions, and final recommendation directly from the new artifacts.

This is very close. The core prototype and evidence process are much stronger now; the remaining work is primarily about bringing the final benchmark truth and report into full alignment with the canonical V2 contract.

**Review status: Request changes.**

jriya186:
Hi Kevin,

Pushed fixes for all five items.

## 1. Grouped skills

Checked every fixture (JSON) with grouped/parenthetical skills. `header_contact_01`, `project_heavy_01`, and `dense_skills_02` were already fine. `multi_col_03` had one leftover: `"Project Management (PMP Certified)"` was split into two skills in the root array. Removed `PMP Certified` — it's a certification, not a separate skill, and it's already listed properly under `CERTIFICATIONS`. `sections` still shows the original wording.

## 2. Golden JSON updates

- `dense_skills_01`: location stays San Francisco, but Santa Barbara is now included in the education entry instead of being left out
- `multi_col_04`: added the skills that were missing (14 items from "Areas of Expertise"). Reran the comparison — this resume went from an unscoreable 0.000 to **+0.387**

## 3. Report count mismatch

Fixed. The fallback list now only shows the 4 resumes that actually fall back. `multi_col_04` was moved into the results table with its real score.

## 4. Docstring fix

Updated `_header_scoped_text()` — it no longer says text is flattened to one line, since that's not what it does anymore after the earlier fix.

## 5. Multi-page note

Added a note (in the code and the report) that the fallback guarantee is per-document, not per-page. Doesn't affect this corpus since everything's single-page. It's just flagged for later.

Reran everything after the JSON changes, results are updated in the report. Let me know if anything else needs a look.

Kevin-TCUS (approved):
Hi Riya,

Thanks for working through the review iterations on this one.

The requested changes have been addressed, and I appreciate the care taken to strengthen both the benchmark corpus and the supporting evidence. The prototype remains appropriately isolated from production, the benchmark integrity is much stronger, and the recommendation to keep the approach experimental is well supported by the current results and documented limitations.

I don't have any further blocking comments. Nice work, and thanks for sticking with this one through multiple review rounds.

Approved.

PR #325, opened 2026-07-13
Define secure resume proxy contract for reviewer dashboard (issue 310)

### LINKED ISSUE #310: [API] Define secure resume proxy contract for reviewer dashboard
## Goal

Define the secure resume access contract for reviewer dashboard users.

The reviewer dashboard should access resume files by `submission_id`, not by exposing raw Drive paths or relying on email/name/filename lookups.

## Background

Libelle stores volunteer intake data in `submissions`, including resume file reference information when a resume is uploaded.

For v0.4, resume access should follow the same trust principles as the rest of the pipeline:

- `submission_id` is the canonical key
- reviewer access is protected
- missing or broken resume references are visible, not silent
- raw volunteer data is not overwritten or inferred from derived data

## Proposed endpoint

`GET /resumes/{submission_id}`

The exact response shape can be finalized in implementation, but the contract should be documented first.

## Requirements

- Resume lookup uses `submission_id` only.
- The endpoint requires reviewer access.
- The endpoint does not trust client-provided file paths.
- Missing resume should return an honest no-resume response.
- Broken Drive reference should return a visible degraded/error response.
- Access failures should not make the submission disappear from `/snapshot`.

## Acceptance criteria

- A backend contract is documented for secure resume access.
- Tests cover valid resume lookup by `submission_id`.
- Tests cover no-resume submission.
- Tests cover missing/broken Drive reference.
- Tests cover unauthorized access, if auth test helpers exist.
- The implementation does not expose raw Drive paths to unauthenticated users.
- The implementation does not look up resumes by email, name, or filename alone.

## Out of scope

- Public resume access.
- Full file preview UI.
- Replacing Google Drive storage.
- Role-based permissions beyond current reviewer access.

briannadelariva:
Closes #310.

The proxy endpoint itself already exists (`GET /resumes/{submission_id}` with `require_internal_actor` and backend-mediated Drive access), so this PR does what the issue asks for on top of it: **documents the contract first**, and closes the one behavior gap found against it.

## Docs
- New **Reviewer Endpoints** section in `docs/api-spec.md` specifying the contract: `submission_id`-only lookup (no email/name/filename lookups, no client-provided paths), internal-actor auth via the same Cloudflare Access derivation as ops writes, audit logging of every served/denied attempt, read-only guarantee (access failures never affect `/snapshot` visibility), response headers, and the full error-code table.

## Behavior fix
- Previously, if the Drive metadata lookup succeeded but the **download itself** failed (revoked token, file deleted mid-request, Drive API error), the exception propagated as an opaque 500. Per the issue's "broken Drive reference should return a visible degraded/error response", this now returns `502 RESUME_FETCH_FAILED` with the denial audit-logged. One narrow try/except; no other behavior changes.

## Acceptance criteria mapping
- Backend contract documented ✔ (api-spec.md)
- Valid lookup by `submission_id` ✔ (existing test)
- No-resume submission ✔ (existing test + new test pinning `resume_status=failed` → honest `RESUME_NOT_AVAILABLE`)
- Missing/broken Drive reference ✔ (existing `RESUME_FILE_NOT_FOUND` test + new `RESUME_FETCH_FAILED` download-failure test)
- Unauthorized access ✔ (existing 401 test)
- No raw Drive paths exposed ✔ (response is proxied PDF bytes + `X-Submission-Id` only; documented)
- No email/name/filename-alone lookups ✔ (service resolves the filename from the submission's own row; documented)

## Testing
Full backend suite: 272 passed.

### DIFF
--- a/backend/services/resume_access_service.py
+++ b/backend/services/resume_access_service.py
@@ -72,13 +72,21 @@ def get_mediated_resume(submission_id: str, actor: str) -> MediatedResume:
                 message="Resume metadata exists, but the file is unavailable.",
             )
 
-        content = download_file(str(drive_file["id"]).strip())
+        drive_file_id = str(drive_file["id"]).strip()
+        try:
+            content = download_file(drive_file_id)
+        except Exception as exc:
+            raise ResumeAccessError(
+                status_code=502,
+                code="RESUME_FETCH_FAILED",
+                message="Resume file exists, but could not be retrieved from storage.",
+            ) from exc
         _log_resume_access(
             submission_id=normalized_submission_id,
             actor=normalized_actor,
             outcome="served",
             filename=filename,
-            drive_file_id=str(drive_file["id"]).strip(),
+            drive_file_id=drive_file_id,
         )
         return MediatedResume(
             submission_id=normalized_submission_id,
--- a/backend/tests/test_resume_access_service.py
+++ b/backend/tests/test_resume_access_service.py
@@ -85,3 +85,55 @@ def test_get_mediated_resume_rejects_missing_drive_file(monkeypatch) -> None:
 
     assert exc_info.value.status_code == 404
     assert exc_info.value.code == "RESUME_FILE_NOT_FOUND"
+
+
+def test_get_mediated_resume_surfaces_download_failure_as_fetch_error(
+    monkeypatch, caplog
+) -> None:
+    monkeypatch.setattr(
+        "storage.sheets_repo.load_submission_records",
+        lambda: {
+            "sub_001": {
+                "submission_id": "sub_001",
+                "resume_filename": "sub_001-resume.pdf",
+                "resume_status": "uploaded",
+            }
+        },
+    )
+    monkeypatch.setattr(
+        "storage.drive_repo.find_pdf_by_name",
+        lambda filename: {"id": "drive-file-1", "name": filename},
+    )
+
+    def failing_download(file_id):
+        raise RuntimeError("drive api unavailable")
+
+    monkeypatch.setattr("storage.drive_repo.download_file", failing_download)
+
+    with caplog.at_level(logging.INFO, logger=resume_access_service.__name__):
+        with pytest.raises(ResumeAccessError) as exc_info:
+            get_mediated_resume("sub_001", "reviewer@example.org")
+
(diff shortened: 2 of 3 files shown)


Kevin-TCUS (changes requested):
Brianna, this is very close and the direction is strong.

What looks good:

- The endpoint contract is correctly framed as backend-mediated reviewer resume access, not direct Drive exposure.
- Lookup is by `submission_id` only.
- The service resolves the resume filename from the submission’s own row rather than trusting email, name, filename, or client-provided paths.
- No-resume cases are handled honestly through `RESUME_NOT_AVAILABLE`.
- Broken Drive metadata is separated from Drive download failure.
- Drive download failure is converted into `RESUME_FETCH_FAILED` instead of leaking through as an opaque storage exception.
- Access failures do not modify source tabs and do not affect `/snapshot` visibility.
- The tests cover valid lookup, missing submission, no uploaded resume, missing Drive file, Drive fetch failure, and failed upload status.
- The API docs clearly state the response headers and error-code contract.

One small clarification before approval:

The API spec currently says every access attempt, served or denied, is logged.

Based on `backend/services/resume_access_service.py`, served requests and service-level denials are logged inside `get_mediated_resume()`. But route-level denials such as missing Cloudflare/internal actor identity may be handled before `get_mediated_resume()` runs.

Could you either:

1. adjust the docs to say “every mediated access attempt that reaches the resume access service is logged,” or
2. confirm/add route-level logging for unauthorized attempts as well?

I do not think this needs a redesign. I just want the audit/logging contract to be precise, especially because this endpoint is security-sensitive.

After that clarification, I think this is approve-ready.

Kevin-TCUS:
Thank you, @briannadelariva . This PR identified and documented an important gap in the resume proxy contract, especially the need to distinguish a Drive retrieval failure from a normal no-resume state.

A broader implementation for #310 has now been completed in #339 using drive_file_id as the authoritative storage reference rather than filename-based Drive lookup. That PR also preserves the retrieval-failure distinction through an explicit storage/retrieval error response.

Because #339 supersedes the underlying lookup contract, I’m going to close this PR rather than merge two overlapping implementations. Your work was useful in shaping the final error and documentation contract—thank you for catching that gap.

PR #339, opened 2026-07-22
Define secure resume proxy contract for reviewer dashboard (for issue 310)

### LINKED ISSUE #310: [API] Define secure resume proxy contract for reviewer dashboard

jliu233:
## Summary

Closes #310.

This PR defines and implements the secure backend resume access contract for reviewer dashboard users.

The reviewer dashboard now accesses resumes through:

```http
GET /resumes/{submission_id}
```

The endpoint uses `submission_id` as the canonical lookup key, requires authenticated reviewer/internal actor access, and proxies the PDF from Drive without exposing raw Drive paths or trusting client-provided file references.

### Secure resume proxy contract

- Documented `GET /resumes/{submission_id}` in `docs/api-spec.md`.
- Defined response behavior for:
  - successful PDF proxy
  - no-resume submissions
  - missing submissions
  - broken Drive references
  - unauthorized access
- Clarified that resume access is separate from `/snapshot`, so resume failures do not remove a submission from reviewer visibility.

### Backend implementation

- Updated resume access mediation to:
  - resolve the submission row by `submission_id`
  - read the stored Drive file ID from that row
  - download the Drive file by stored file ID
  - use `resume_filename` only as a display/download filename
- Removed filename-based Drive lookup from the resume access path.
- Added explicit degraded handling for broken Drive references with `RESUME_REFERENCE_BROKEN`.

### Submission schema and snapshot behavior

- Added storage-backed resume reference fields to the submissions schema:
  - `drive_file_id`
  - `drive_file_url`
- Updated base-row writes to persist those fields.
- Kept storage-only Drive references out of `/snapshot`.
- Updated architecture docs to make `drive_file_id` the resume file reference source of record.

### Tests

Added or updated tests for:

- valid resume lookup by `submission_id`
- no-resume submission
- missing/broken Drive reference
- unauthorized route access
- regression protection against filename-based lookup
- intake persistence of Drive file references
- `/snapshot` exclusion of Drive file IDs/URLs

## Security Notes

This implementation does not trust client-provided:

- Drive paths
- Drive URLs
- Drive file IDs
- filenames
- volunteer names
- email addresses

The client supplies only `submission_id`. Reviewer identity is derived server-side from the existing internal actor mechanism.

The PDF response does not expose raw Drive paths, Drive URLs, or Drive file IDs.

## Operational Note

Because this adds `drive_file_id` and `drive_file_url` to the `submissions` schema, the live Google Sheet must be migrated before startup schema validation will pass in deployed environments.

## Verification

Ran:

```bash
backend/.venv/bin/python -m compileall -q backend
```

Result: passed.


Kevin-TCUS (changes requested):
Hi Jiaqi — this is a strong architectural direction for #310.

Moving resume access from filename search to:

`submission_id → immutable submission row → drive_file_id → mediated PDF`

is the right long-term contract for Libelle. It keeps storage identity server-side, prevents the client from selecting a Drive artifact, preserves submission visibility when resume access fails, and gives the frontend a stable reviewer endpoint. I also like the regression test that prevents filename-based lookup from being reintroduced.

I reviewed this together with the current staging sheet and the field-ownership, precedence, and API contracts. Since the existing staging rows are only test data, we do not need a legacy filename fallback or historical backfill. We can reset those rows and establish the cleaner contract now.

I am requesting a few bounded changes before merge:

1. **Keep `drive_file_id`, but remove `drive_file_url`.**

   `drive_file_id` is important because it gives Libelle a stable, server-controlled reference to the exact stored resume. `drive_file_url` does not add meaningful identity or provenance, is not used by the secure proxy, can become stale, and creates ambiguity about whether Drive URLs are internal or public.

   Please remove `drive_file_url` from the sheet schema, field-ownership constants, intake persistence and return values, snapshot exclusions, tests, and documentation. The public upload response should return Libelle concepts such as `submission_id`, `resume_status`, and confirmation—not a Google Drive URL.

2. **Preserve `failed` versus `missing`.**

   `_resume_reference_from_submission()` currently maps every non-`uploaded` status to `RESUME_NOT_AVAILABLE`.

   These states have different meanings:

   * `missing`: the volunteer did not provide a resume
   * `failed`: the volunteer provided one, but Libelle failed to store it

   Please return a distinct stable error for `resume_status="failed"` and add a regression test. An upload failure should not be represented to reviewers as though no resume was supplied.

3. **Prove the route-level authorization boundary.**

   The service tests pass an actor directly, but I do not see a route-level test proving that a request without trusted internal-actor identity receives `401 INTERNAL_ACTOR_REQUIRED` before Drive access.

   If that coverage already exists elsewhere, please link it in the PR. Otherwise, please add a route test confirming the resume service or Drive repository is not called when reviewer identity is missing.

4. **Do not classify every Drive exception as a broken reference.**

   The broad `except Exception` currently maps timeouts, credential failures, temporary Drive outages, and unexpected errors to `RESUME_REFERENCE_BROKEN`.

   A full typed Drive exception hierarchy is not necessary here, but generic retrieval failures should use a neutral code such as `RESUME_RETRIEVAL_FAILED` or `RESUME_STORAGE_UNAVAILABLE`. Reserve `RESUME_REFERENCE_BROKEN` for a blank stored ID or a confirmed missing/inaccessible file.

5. **Align the filename documentation with the implementation.**

   The ownership contract says deterministic `{submission_id}` naming, while intake currently prefixes the original filename with the submission ID.

   Either behavior is acceptable because the filename is display-only, but the documented invariant should match the actual behavior.

For staging, once the final schema is settled, we can update the live sheet headers, clear the existing test rows, restart the backend, and create fresh submissions through the updated intake path. No migration script or legacy fallback is needed.

The core proxy design is good and should remain. These are focused contract, trust-boundary, and state-semantics corrections—not a request to redesign the implementation.

jliu233:
Thank you very much for all your feedback, Kevin! I’ve addressed the bounded changes while keeping the core proxy design intact.

Updates made:

- Removed `drive_file_url` from the backend contract:
  - sheet schema
  - field ownership/state constants
  - intake persistence and return payloads
  - public upload response
  - snapshot exclusions
  - tests
  - API/architecture docs

- Kept `drive_file_id` as the only stored Drive reference used by the secure proxy.

- Preserved `missing` vs `failed` semantics:
  - `missing` returns `RESUME_NOT_AVAILABLE`
  - `failed` now returns `RESUME_UPLOAD_FAILED`
  - added regression coverage for the failed-upload case

- Strengthened route-level auth coverage:
  - unauthenticated `GET /resumes/{submission_id}` now has a test proving it returns `401 INTERNAL_ACTOR_REQUIRED` before the resume service is called

- Refined Drive retrieval error mapping:
  - blank stored Drive ID remains `RESUME_REFERENCE_BROKEN`
  - confirmed 403/404-style missing or inaccessible Drive file remains `RESUME_REFERENCE_BROKEN`
  - generic retrieval/storage failures now return `RESUME_RETRIEVAL_FAILED`

- Aligned filename documentation with the implementation:
  - docs now describe the current `submission_id`-prefixed original filename behavior
  - filename is documented as display-only, not part of resume lookup identity

Also verified that `drive_file_url` no longer appears in backend/docs, and `compileall` passes locally.

Kevin-TCUS (approved):
Thanks, Jiaqi. I reviewed the revised implementation and tests, and the requested changes are resolved.

The final contract is much stronger:

* drive_file_id is the sole authoritative storage reference.
* drive_file_url has been removed from persistence, schema, public responses, tests, and documentation.
* filename-based Drive lookup is eliminated and protected by regression coverage.
* missing, failed, broken-reference, and general retrieval-failure states remain distinct.
* unauthenticated requests are rejected before the resume access service is called.
* storage-only identity remains outside /snapshot.

This is the right architectural direction for v0.4 and materially strengthens the provenance and trust boundary around resume access.

One non-blocking follow-up: _download_error_response() currently maps both Drive 403 and 404 to RESUME_REFERENCE_BROKEN. Since some 403 responses may represent rate limits, credentials, or broader provider availability rather than a broken file reference, we should consider mapping only confirmed 404 responses to RESUME_REFERENCE_BROKEN and allowing generic 403 responses to fall through to RESUME_RETRIEVAL_FAILED.

Approved.

jliu233:
Thank you very much for your approval, Kevin!:)

I agree on the non-blocking follow-up. Mapping `404` to `RESUME_REFERENCE_BROKEN` is clearly right, but generic Drive `403` can mean several things beyond a bad file reference, including auth scope, rate limiting, or provider-side access behavior. Treating generic `403` as `RESUME_RETRIEVAL_FAILED` would keep the contract more honest unless we later inspect a typed Drive error reason that confirms the file itself is inaccessible.

I’ll leave the approved contract as-is for this PR:)

PR #351, opened 2026-08-04
Add public intake file validation layer (for issue 336)

### LINKED ISSUE #336: [Backend] Add public intake file validation layer
## Context

Libelle v0.4 is intended to provide a trusted intake pipeline from volunteer submission through reviewer action.

The public intake endpoint currently accepts resume uploads that may later be written to storage or passed into parser processing. Before that happens, the backend should reject unsupported, empty, oversized, or malformed files using one consistent validation layer.

Validation should happen before Google Drive persistence and before parser execution.

## Goal

Add a centralized backend file-validation layer for public resume uploads.

The validator should provide predictable, frontend-safe errors while preventing rejected files from entering downstream storage or processing.

## Requirements

Validate uploaded resume files for:

- accepted file type
- allowed filename extension
- declared MIME type
- extension and MIME-type disagreement
- maximum configured file size
- empty file content
- malformed or unreadable PDF content
- missing filename or missing upload payload

For this issue, the supported production format should remain PDF unless the existing intake contract explicitly supports additional formats.

Validation should run before:

- Google Drive upload
- submissions persistence that claims a resume was accepted
- parser invocation
- resolver invocation

Use one reusable validation path rather than duplicating checks across endpoint, storage, and parser code.

## Error Contract

Rejected files should return a stable client-facing response that includes:

- an appropriate HTTP status
- a machine-readable error code
- a short reviewer-safe or volunteer-safe message
- no internal filesystem, Drive, stack-trace, or parser details

Suggested error categories may include:

- `missing_file`
- `empty_file`
- `unsupported_file_type`
- `file_too_large`
- `invalid_pdf`

The final names should follow existing API conventions.

## Acceptance Criteria

- A valid PDF within the configured size limit is accepted.
- An empty upload is rejected.
- A file above the configured size limit is rejected.
- A non-PDF upload is rejected.
- A file renamed with a `.pdf` extension but not containing readable PDF content is rejected.
- MIME-type and extension disagreement is handled explicitly.
- Rejected files are not uploaded to Drive.
- Rejected files are not sent to the parser.
- Rejected files do not create a misleading successful intake state.
- Validation errors use a stable frontend-safe response shape.
- Maximum file size is defined in configuration rather than buried as an unexplained literal.
- Existing valid intake behavior continues to work.
- Automated tests cover accepted and rejected cases.

## Manual Verification

Verify at minimum:

- valid PDF
- empty PDF
- oversized PDF
- text or image file renamed to `.pdf`
- non-PDF file
- MIME-type mismatch
- missing upload

## Non-Goals

- Malware scanning
- Antivirus integration
- OCR validation
- Parser quality improvements
- Supporting additional document formats
- Resume-content moderation
- Changes to matching or reviewer workflow behavior

jliu233:
## Summary

Closes #336.

Adds a centralized backend validation layer for public intake resume uploads so unsupported, empty, oversized, mismatched, or unreadable files are rejected before they can enter Drive storage, submission persistence as an accepted resume, or parser/resolver processing.

## Changes

- Added `backend/services/intake_file_validation.py` as the single reusable resume upload validation path.
- Validates uploaded resume files for:
  - missing file payload
  - missing filename
  - empty content
  - unsupported filename extension
  - declared MIME type mismatch
  - configured maximum file size
  - invalid or unreadable PDF bytes
  - PDFs with no extractable text
- Keeps production support limited to PDF:
  - allowed extension: `.pdf`
  - allowed MIME types: `application/pdf`, `application/x-pdf`
- Uses `MAX_PDF_MB` from configuration for the file size limit.
- Updates `finalize_submission()` so validation runs before:
  - Google Drive upload
  - base submission row write claiming an uploaded resume
  - parser background task scheduling
- Preserves the existing optional-resume contract:
  - no file at all still records `resume_status="missing"`
  - a present-but-invalid upload is rejected
- Returns stable frontend-safe error payloads through the existing intake error shape.

## Error Codes

New/updated public intake file validation codes include:

- `MISSING_FILE`
- `EMPTY_FILE`
- `UNSUPPORTED_FILE_TYPE`
- `MIME_TYPE_MISMATCH`
- `FILE_TOO_LARGE`
- `INVALID_PDF`

Oversized uploads return `413`; other file validation failures return `400`.

## Tests Added / Updated

Expanded intake service and route coverage for:

- valid readable PDF accepted
- empty upload rejected
- configured size limit enforced
- non-PDF extension rejected
- `.pdf` filename with mismatched MIME rejected
- `.pdf` filename with non-PDF bytes rejected
- missing filename rejected
- filename metadata without payload rejected
- password-protected/unreadable PDF rejected as `INVALID_PDF`
- invalid uploads do not call Drive upload
- invalid uploads do not write a misleading successful submission row
- invalid uploads do not schedule parser work
- no-resume submission still succeeds without parser work

## Verification

Ran:

```bash
backend/.venv/bin/python -m compileall backend/services/intake_file_validation.py backend/services/intake_service.py backend/api/routes/intake.py backend/tests/test_intake_service.py backend/tests/test_intake_route_rate_limit.py
```

Also manually exercised validator behavior for:

- valid generated PDF
- empty file
- MIME mismatch
- fake PDF bytes
- missing payload

And manually exercised the FastAPI intake route to confirm an invalid renamed PDF returns:

```text
400 INVALID_PDF
```

with no Drive, Sheets, or parser calls.

### DIFF
--- a/backend/api/routes/intake.py
+++ b/backend/api/routes/intake.py
@@ -88,10 +88,11 @@ async def upload_volunteer_application(
                 },
                 headers={"Retry-After": str(decision.retry_after_seconds)},
             )
-        pdf_bytes = await file.read() if file and file.filename else None
+        pdf_bytes = await file.read() if file else None
         result = finalize_submission(
             pdf_bytes=pdf_bytes,
-            original_filename=file.filename if file and file.filename else None,
+            original_filename=file.filename if file else None,
+            content_type=file.content_type if file else None,
             normalized=normalized,
             linkedin_url=linkedin_url,
             github_url=github_url,
--- a/backend/services/intake_file_validation.py
+++ b/backend/services/intake_file_validation.py
@@ -1,0 +1,129 @@
+"""Central validation for public intake resume uploads."""
+
+from dataclasses import dataclass
+from pathlib import PurePath
+from typing import Optional
+
+from config import MAX_PDF_MB
+from services.pdf_text_extraction import (
+    PasswordProtectedPDFError,
+    extract_text_from_pdf_bytes,
+)
+
+
+ALLOWED_PDF_EXTENSIONS = {".pdf"}
+ALLOWED_PDF_MIMES = {"application/pdf", "application/x-pdf"}
+
+
+@dataclass(frozen=True)
+class ValidatedResumeUpload:
+    pdf_bytes: bytes
+    original_filename: str
+    content_type: str
+    extracted_text: str
+
+
+class ResumeFileValidationError(Exception):
+    def __init__(self, code: str, message: str, status_code: int = 400):
+        super().__init__(message)
+        self.code = code
+        self.message = message
+        self.status_code = status_code
+
+
+def _normalized_content_type(content_type: Optional[str]) -> str:
+    return (content_type or "").split(";", 1)[0].strip().lower()
+
+
+def _filename_extension(filename: str) -> str:
+    return PurePath(filename.strip()).suffix.lower()
+
+
+def validate_resume_upload(
+    *,
+    filename: Optional[str],
+    content_type: Optional[str],
+    file_bytes: Optional[bytes],
+) -> ValidatedResumeUpload:
+    """Validate and pre-extract an optional public intake PDF upload."""
+
+    if file_bytes is None:
+        raise ResumeFileValidationError(
+            "MISSING_FILE",
+            "No resume file was received.",
+            status_code=400,
+        )
+
+    original_filename = (filename or "").strip()
+    if not original_filename:
+        raise ResumeFileValidationError(
+            "MISSING_FILE",
+            "No resume filename was received.",
+            status_code=400,
+        )
+
+    if not file_bytes:
+        raise ResumeFileValidationError(
+            "EMPTY_FILE",
+            "The uploaded resume file is empty.",
+            status_code=400,
+        )
+
+    extension = _filename_extension(original_filename)
+    if extension not in ALLOWED_PDF_EXTENSIONS:
+        raise ResumeFileValidationError(
+            "UNSUPPORTED_FILE_TYPE",
+            "Only PDF resume files are supported.",
+            status_code=400,
+        )
+
+    normalized_content_type = _normalized_content_type(content_type)
+    if normalized_content_type and normalized_content_type not in ALLOWED_PDF_MIMES:
+        raise ResumeFileValidationError(
+            "MIME_TYPE_MISMATCH",
+            "The resume filename and declared file type do not agree.",
+            status_code=400,
+        )
+
+    if len(file_bytes) > MAX_PDF_MB * 1024 * 1024:
+        raise ResumeFileValidationError(
+            "FILE_TOO_LARGE",
+            f"Resume PDF must be {MAX_PDF_MB}MB or smaller.",
+            status_code=413,
+        )
+
+    if b"%PDF-" not in file_bytes[:1024]:
+        raise ResumeFileValidationError(
+            "INVALID_PDF",
+            "The uploaded resume is not a readable PDF.",
+            status_code=400,
+        )
+
+    try:
+        extracted_text = extract_text_from_pdf_bytes(file_bytes)
+    except PasswordProtectedPDFError:
+        raise ResumeFileValidationError(
+            "INVALID_PDF",
+            "The uploaded resume is not a readable PDF.",
+            status_code=400,
+        )
+    except Exception:
+        raise ResumeFileValidationError(
+            "INVALID_PDF",
+            "The uploaded resume is not a readable PDF.",
+            status_code=400,
+        )
+
+    if not extracted_text.strip():
+        raise ResumeFileValidationError(
+            "INVALID_PDF",
+            "The uploaded resume is not a readable PDF.",
+            status_code=400,
+        )
+
+    return ValidatedResumeUpload(
+        pdf_bytes=file_bytes,
+        original_filename=original_filename,
+        content_type=normalized_content_type or "application/pdf",
+        extracted_text=extracted_text,
+    )
--- a/backend/services/intake_service.py
+++ b/backend/services/intake_service.py
@@ -5,12 +5,15 @@ import traceback
 import uuid
 from typing import Any, Dict, List, Optional, Union
 
-from config import MAX_PDF_MB
+from services.intake_file_validation import (
+    ResumeFileValidationError,
+    ValidatedResumeUpload,
+    validate_resume_upload,
+)
 from storage.drive_repo import upload_pdf
 from storage.sheets_repo import append_error_row, write_base_row
 
 
-ALLOWED_PDF_MIMES = {"application/pdf", "application/x-pdf"}
 EMAIL_IN_TEXT_RE = re.compile(r"[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}")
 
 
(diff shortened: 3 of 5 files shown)


Kevin-TCUS (changes requested):
Hi Jiaqi,

Nice work on this one. I think centralizing the validation into a single reusable service is the right design, and it cleanly satisfies the intent of #336:

- validation happens before Drive upload, persistence, and parser execution;
- the validator provides one reusable path instead of duplicating checks;
- invalid uploads return stable client-facing error codes;
- valid "no resume" submissions continue to follow the existing optional-resume contract;
- the test coverage is substantially stronger than before.

I just have one requested change before merging.

### Requested change

The validator currently checks whether the uploaded bytes contain the PDF signature using:

```python
if b"%PDF-" not in file_bytes[:1024]:
```

This is a reasonable fast sanity check, but it isn't really necessary once the file is immediately passed to `extract_text_from_pdf_bytes()`, which already performs the authoritative validation and is what ultimately determines whether the upload is a readable PDF.

Relying on the parser itself as the single source of truth avoids maintaining two partially-overlapping validation rules and reduces the chance of future disagreement if unusual but valid PDFs are encountered.

I'd recommend either removing the header check entirely or clearly documenting that it is only an inexpensive early rejection rather than the actual PDF validation step.

---

Overall I think this is a solid implementation and moves the intake pipeline in the right direction. Once the validation path is simplified or clarified, I'm happy to approve.

Kevin-TCUS (changes requested):
Hi @jliu233,

This is a strong implementation. Centralizing validation before Google Drive upload, submission persistence, and parser scheduling is exactly the right architecture for this stage of the intake pipeline. The new validation layer is clean, the error contract is stable and frontend-friendly, and the test coverage around invalid uploads is thorough. I also like the early PDF magic-byte check before handing arbitrary bytes to the PDF parser.

Before merging, I think there is one edge case worth addressing around optional resume uploads.

### Optional Resume Edge Case

`finalize_submission()` currently detects "no resume provided" using:

```python
if pdf_bytes is None and not original_filename and not content_type:
```

However, browsers and FastAPI can construct an `UploadFile` object even when a user leaves an optional `<input type="file">` empty.

In that situation, the request may look something like:

```python
pdf_bytes = b""
original_filename = ""
content_type = "application/octet-stream"
```

Because `b""` is not `None`, the current guard may skip the intended "no resume provided" path and instead forward the upload into `validate_resume_upload()`, which would correctly reject it as `MISSING_FILE`.

If that occurs, a volunteer intentionally submitting without a resume could receive a `400` instead of following the existing optional-resume flow.

### Requested Change

Could you either:

- update the missing-resume guard so an empty upload is treated the same as no upload, **or**
- add an integration test that exercises the browser/FastAPI empty-file submission path and confirms the current implementation already behaves correctly?

I mainly want to make sure the optional-resume contract continues to hold for real browser submissions, not just requests that omit the file field entirely.

Once that edge case is addressed (or demonstrated by a test not to be an issue), I'm happy to approve. Nice work on this overall.

jliu233:
Hi Kevin, thank you for all your feedback! I addressed both requested changes:

- Removed the `%PDF-` header sanity check so `extract_text_from_pdf_bytes()` is now the single source of truth for readable PDF validation.
- Updated the optional-resume path so an empty browser/FastAPI file part, `b""` with no filename and a default content type, is treated the same as no resume provided.
- Added coverage for that edge case at both the service level and the FastAPI multipart route level.

Verified with:

```bash
backend/.venv/bin/python -m pytest backend/tests/test_intake_service.py backend/tests/test_intake_route_rate_limit.py
```

All focused tests pass: `20 passed`.

Kevin-TCUS (changes requested):
Hi Jiaqi,

Thank you for the revision. The optional-resume browser edge case is now handled correctly, and the new service- and route-level regression tests are exactly what I was hoping to see. I also like that PDF parsing is now the single authoritative content-validation path rather than maintaining a separate magic-byte rule.

I noticed two related failure-semantics details that I think we should tighten before merging.

### 1. Distinguish an invalid PDF from a valid PDF with no extractable text

The validator currently maps both parser/open failures and this case:

    if not extracted_text.strip():

to:

    INVALID_PDF
    "The uploaded resume is not a readable PDF."

A structurally valid scanned/image-only PDF can open successfully while producing no extractable text. Since OCR is explicitly outside the scope of #336, I do not think we need to add OCR or redesign the intake path here, but I would like the contract to preserve the distinction.

Please either:

- use a separate frontend-safe error code/message for a valid PDF that contains no extractable text, such as `NO_EXTRACTABLE_TEXT`; or
- explicitly document that `INVALID_PDF` intentionally includes PDFs that cannot be consumed by Libelle's current text-extraction path.

My preference is the separate state/error code because it preserves more accurate evidence for future OCR or parser work.

### 2. Do not convert every unexpected extraction exception into `INVALID_PDF`

The validator currently has:

    except Exception:
        raise ResumeFileValidationError(
            "INVALID_PDF",
            "The uploaded resume is not a readable PDF.",
            status_code=400,
        )

This can turn an unexpected runtime, dependency, or infrastructure failure into a client-facing claim that the volunteer's PDF is invalid.

Please catch the known PDF/input exceptions that actually mean the document is malformed, unreadable, or password-protected, and allow unexpected exceptions to follow the normal server-side failure/observability path instead of being mislabeled as a 400 validation failure.

The important contract is:

- known bad/unsupported document → stable frontend-safe validation error;
- unexpected system failure → operational/server failure, not a claim that the source file was invalid.

Everything else in the revision looks strong. The empty browser file-part behavior is now correct, the new regression coverage is good, and validation still happens before Drive persistence or parser scheduling.

Once these failure semantics are tightened, I think this is ready.

jliu233:
Hi Kevin, thank you for this feedback! I agree with both points, and I’ve tightened the failure semantics accordingly.

I changed the validator so a PDF that opens successfully but produces no extracted text now returns a distinct frontend-safe validation error:

`NO_EXTRACTABLE_TEXT`
`The uploaded resume PDF does not contain extractable text.`

I also removed the broad `except Exception` mapping to `INVALID_PDF`. The validator now only converts known PDF/input failures, including PyMuPDF document data errors and password-protected PDFs, into the stable `INVALID_PDF` validation response. Unexpected extraction/runtime failures are allowed to propagate through the normal server-side error path instead of being presented as a client-side invalid-document claim.

I added service-level coverage for both cases, plus a route-level regression for the `NO_EXTRACTABLE_TEXT` response to confirm Drive persistence and parser scheduling are still skipped.

Kevin-TCUS (changes requested):
Hi Jiaqi,

Thank you again for the revision. I reviewed the full files closely, and the previous validation/failure-semantics concerns are addressed well.

`NO_EXTRACTABLE_TEXT` is now distinct from `INVALID_PDF`, unexpected extraction failures correctly propagate through the server-side failure path, and the optional-resume and rejected-upload regression coverage looks good.

I found one integration issue we should fix before merging, though, because #351 and #361 are currently evolving the same Drive upload contract.

### Align #351 with the finalized Drive upload contract from #361

This branch currently still expects `upload_pdf()` to return three values:

```python
drive_file_id, _, resume_filename = upload_pdf(
    validated_resume.pdf_bytes,
    submission_id,
    validated_resume.original_filename,
)
```

and `test_intake_service.py` mocks that same legacy contract:

```python
return (
    "drive-file-id",
    "https://drive.example/view",
    f"{submission_id}_{original_filename}",
)
```

However, #361 finalizes the v0.4 storage contract by removing `webViewLink` / `drive_file_url` and changing `upload_pdf()` to return only:

```text
(drive_file_id, resume_filename)
```

Since that is the contract we intend to carry forward, #351 should use the same interface before merge.

Please update the intake call to:

```python
drive_file_id, resume_filename = upload_pdf(
    validated_resume.pdf_bytes,
    submission_id,
    validated_resume.original_filename,
)
```

and update the corresponding test mock to return the same two-value tuple.

This is important because otherwise, after #361 lands, the successful-upload path in #351 would attempt to unpack a two-value return into three variables.

I don't see any additional validation-design changes needed. The previous review items are addressed; this is just making the two open v0.4 PRs agree on the finalized storage boundary.

Once this is aligned, I think #351 is ready.

Kevin-TCUS (approved):
Thanks for the revision, Jiaqi.

I reviewed the current files again after the merge-conflict resolution and rechecked the implementation against the previous review comments and the finalized v0.4 storage contract.

The remaining integration issue is resolved: #351 now uses the same Drive upload interface established by #361, with `upload_pdf()` returning `(drive_file_id, resume_filename)`, and the corresponding test double matches that production signature.

I also rechecked the validation boundary and failure semantics:

- rate limiting occurs before PDF content parsing and extraction;
- the centralized validator is the authoritative resume-validation path;
- malformed/unreadable and password-protected PDFs produce bounded client-safe validation errors;
- a valid PDF with no extractable text remains distinct as `NO_EXTRACTABLE_TEXT`;
- unexpected extraction/runtime failures are not mislabeled as invalid user input;
- an empty browser file part preserves the optional no-resume flow;
- a present-but-empty file is rejected explicitly;
- rejected uploads do not reach Drive persistence, accepted submission persistence, or parser scheduling; and
- the successful upload path preserves the same `submission_id` through Drive and Sheets.

The revised tests cover the important accepted, rejected, optional-resume, failure, and integration paths well.

I don't see another blocking issue here. This now aligns well with the intent of #336 and the v0.4 trusted-intake contract.

Nice work working through the review iterations carefully.

**Approved.**

PR #353, opened 2026-08-05
Harden intake client-IP trust behind Cloudflare (for issue 337)

### LINKED ISSUE #337: [Backend/Security] Harden intake client-IP trust behind Cloudflare
## Context

Basic public-intake rate limiting landed in #245 through PR #262.

The current implementation uses `cf-connecting-ip`, `x-forwarded-for`, and the socket client address as possible client-IP sources. Before relying on this in staging or production, Libelle should explicitly define which proxy headers are trusted and under what deployment conditions.

## Goal

Harden client-IP resolution for intake rate limiting so arbitrary forwarded headers cannot be used to spoof or bypass per-IP limits.

## Requirements

- Define the trusted proxy boundary for local, staging, and production environments.
- Prefer `cf-connecting-ip` only when the request is known to have passed through the configured Cloudflare boundary.
- Do not trust arbitrary `x-forwarded-for` values from direct public requests.
- Fall back safely to the socket client address when trusted proxy conditions are not met.
- Keep the existing per-IP, per-email, and global rate-limit behavior unchanged.
- Document expected origin-access restrictions for production deployment.
- Add tests covering trusted and untrusted forwarded-header scenarios.

## Acceptance Criteria

- Direct requests cannot select their own client identity through arbitrary forwarding headers.
- Requests through the configured Cloudflare path resolve the intended client IP.
- Existing rate-limit response behavior remains unchanged.
- Existing rate-limit tests continue to pass.
- Proxy trust behavior is documented for local, staging, and production deployment.

## Non-Goals

- Replacing the in-memory limiter
- Distributed rate limiting
- Cloudflare WAF rules
- CAPTCHA
- New abuse-detection logic

jliu233:
## Summary

Closes #337.

This PR hardens public intake client-IP resolution so rate limiting no longer trusts arbitrary forwarding headers from direct requests. The intake route now resolves client identity through an explicit trusted proxy boundary:

- Trusts `CF-Connecting-IP` only when the socket peer is in `INTAKE_TRUSTED_CLOUDFLARE_PROXY_CIDRS`
- Trusts the leftmost `X-Forwarded-For` value only when the socket peer is in `INTAKE_TRUSTED_FORWARD_PROXY_CIDRS`
- Falls back to the socket client address when the request does not come from a configured trusted proxy
- Ignores malformed forwarded IP values instead of using them as rate-limit keys

The existing per-IP, per-email, and global limiter behavior is unchanged; this only changes how the per-IP key is resolved.

## Changes

- Added env-driven trusted proxy configuration:
  - `INTAKE_TRUSTED_CLOUDFLARE_PROXY_CIDRS`
  - `INTAKE_TRUSTED_FORWARD_PROXY_CIDRS`
- Updated intake `_client_ip()` resolution to:
  - Prefer `CF-Connecting-IP` only from trusted Cloudflare proxy CIDRs
  - Use `X-Forwarded-For` only from explicitly trusted forward proxy CIDRs
  - Fall back safely to `request.client.host`
- Added client-IP trust tests covering:
  - Direct requests with spoofed `CF-Connecting-IP` / `X-Forwarded-For`
  - Trusted Cloudflare boundary resolving the intended volunteer IP
  - Trusted generic forward proxy behavior
  - Invalid forwarded header fallback
- Added deployment documentation for:
  - Local development defaults
  - Staging Cloudflare/nginx trust boundary
  - Production origin-access restrictions

## Deployment Notes

Forwarded headers are now ignored unless the backend socket peer is explicitly trusted.

Recommended staging config for the current Raspberry Pi topology:

```env
INTAKE_TRUSTED_CLOUDFLARE_PROXY_CIDRS=127.0.0.1/32,::1/128
INTAKE_TRUSTED_FORWARD_PROXY_CIDRS=
```

Production should keep origin access restricted so public clients cannot reach the backend or origin nginx directly and inject their own `CF-Connecting-IP` header. The new docs call out expected controls: backend bound to localhost/private interface, Cloudflare Tunnel or equivalent Cloudflare-controlled path, and direct-origin public HTTP/HTTPS blocked where applicable.

## Testing

Ran:

```bash
python3 -m py_compile backend/config.py backend/api/routes/intake.py backend/tests/test_intake_route_rate_limit.py
```

Also ran a lightweight runtime check of the `_client_ip()` trusted/untrusted scenarios using the backend venv.

### DIFF
--- a/backend/api/routes/intake.py
+++ b/backend/api/routes/intake.py
@@ -1,4 +1,5 @@
 import traceback
+from ipaddress import ip_address, ip_network
 from typing import Any, Dict, Optional
 
 from fastapi import APIRouter, BackgroundTasks, File, Form, HTTPException, Request, UploadFile
@@ -9,6 +10,8 @@ from config import (
     INTAKE_RATE_LIMIT_GLOBAL_PER_MINUTE,
     INTAKE_RATE_LIMIT_PER_EMAIL_PER_HOUR,
     INTAKE_RATE_LIMIT_PER_IP_PER_MINUTE,
+    INTAKE_TRUSTED_CLOUDFLARE_PROXY_CIDRS,
+    INTAKE_TRUSTED_FORWARD_PROXY_CIDRS,
 )
 from services.intake_service import IntakeError, finalize_submission, validate_intake
 from services.rate_limit import InMemoryIntakeRateLimiter
@@ -23,6 +26,34 @@ intake_rate_limiter = InMemoryIntakeRateLimiter(
 )
 
 
+def _is_ip_in_cidrs(value: Optional[str], cidrs: list[str]) -> bool:
+    if not value:
+        return False
+
+    try:
+        candidate = ip_address(value)
+    except ValueError:
+        return False
+
+    for cidr in cidrs:
+        try:
+            if candidate in ip_network(cidr, strict=False):
+                return True
+        except ValueError:
+            continue
+    return False
+
+
+def _normalized_ip(value: Optional[str]) -> Optional[str]:
+    if not value:
+        return None
+
+    try:
+        return str(ip_address(value.strip()))
+    except ValueError:
+        return None
+
+
 def _intake_error_to_http(err: IntakeError) -> HTTPException:
     detail: Dict[str, Any] = {"status": "error", "code": err.code}
     if err.fields is not None:
@@ -33,15 +64,21 @@ def _intake_error_to_http(err: IntakeError) -> HTTPException:
 
 
 def _client_ip(request: Request) -> str:
-    cf_ip = request.headers.get("cf-connecting-ip")
-    if cf_ip:
-        return cf_ip.strip()
+    socket_ip = request.client.host if request.client else None
 
-    forwarded_for = request.headers.get("x-forwarded-for")
-    if forwarded_for:
-        return forwarded_for.split(",", 1)[0].strip()
+    if _is_ip_in_cidrs(socket_ip, INTAKE_TRUSTED_CLOUDFLARE_PROXY_CIDRS):
+        cf_ip = _normalized_ip(request.headers.get("cf-connecting-ip"))
+        if cf_ip:
+            return cf_ip
 
-    return request.client.host if request.client else "unknown"
+    if _is_ip_in_cidrs(socket_ip, INTAKE_TRUSTED_FORWARD_PROXY_CIDRS):
+        forwarded_for = request.headers.get("x-forwarded-for")
+        if forwarded_for:
+            forwarded_ip = _normalized_ip(forwarded_for.split(",", 1)[0])
+            if forwarded_ip:
+                return forwarded_ip
+
+    return socket_ip or "unknown"
 
 
 @router.post("/api/upload")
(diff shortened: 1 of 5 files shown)


Kevin-TCUS (changes requested):
Hi @jliu233,

This is a strong security-focused implementation overall. I like the move from blindly trusting forwarded headers to an explicit trusted-proxy boundary, the fail-safe fallback to the socket peer, the handling of malformed forwarded values, and the accompanying documentation describing the intended deployment topology.

I have one requested change before merging.

### Clarify the nginx / Cloudflare trust boundary

The current staging guidance recommends:

```text
INTAKE_TRUSTED_CLOUDFLARE_PROXY_CIDRS=127.0.0.1/32,::1/128
```

Given the documented staging topology:

```text
Cloudflare Tunnel → nginx → FastAPI
```

the backend's socket peer is nginx (loopback), not Cloudflare itself. That means trusting `CF-Connecting-IP` from loopback is only safe if nginx (or the deployment architecture) guarantees that arbitrary clients cannot inject that header.

I'd like the documentation to make that dependency explicit. For example, note that the loopback configuration is only appropriate when at least one of the following is true:

- nginx is reachable only through the Cloudflare-controlled path (for example, via Cloudflare Tunnel with direct origin access blocked), or
- nginx explicitly strips and replaces any inbound `CF-Connecting-IP` header before proxying requests to the backend.

If nginx already performs that sanitization, it would be helpful to document the expected behavior. Otherwise, I'd avoid presenting the loopback CIDR setting by itself as sufficient protection.

Once that clarification is added, this looks good to me. Nice work on a security-sensitive change.

jliu233:
Hi Kevin, thank you very much for your comment! I updated the staging and proxy-trust docs to make the nginx / Cloudflare boundary explicit.

The loopback CIDR example is now presented as conditional on either direct origin access being blocked through the Cloudflare-controlled path, or nginx stripping and replacing any inbound `CF-Connecting-IP` before proxying to FastAPI. That should avoid implying that trusting loopback alone is sufficient protection.

Kevin-TCUS (approved):
Hi Jiaqi,

Thank you for the revision. I reviewed the updated implementation, tests, and deployment documentation.

The clarification around the nginx / Cloudflare trust boundary addresses the concern I raised. The documentation now explicitly describes the intended staging topology:

```text
Cloudflare Tunnel -> nginx -> FastAPI
```

and makes it clear that the backend trusts `CF-Connecting-IP` from the loopback peer only under the documented deployment assumptions—either because direct origin access is blocked through the Cloudflare-controlled path or because nginx strips and replaces any inbound `CF-Connecting-IP` header before proxying requests.

I also rechecked the implementation and am happy with the trust model:

- forwarded headers are ignored unless the socket peer is explicitly trusted;
- `CF-Connecting-IP` and `X-Forwarded-For` have separate trust boundaries;
- malformed forwarded values safely fall back to the socket client address;
- the default configuration remains fail-safe by trusting no forwarded headers.

An end-to-end verification of the deployed Cloudflare → nginx → FastAPI path would still be valuable as part of staging or release verification, but that's an operational concern rather than something this PR needs to prove with unit tests.

This addresses my requested change. Nice work on a security-sensitive improvement. Approved.

PR #354, opened 2026-08-06
Define asynchronous parser job execution model (for issue 338)

### LINKED ISSUE #338: [Architecture/Spike] Define asynchronous parser job execution model
## Context

The v0.4 planning decomposition proposes moving parser execution out of the synchronous public intake request.

Before implementing an asynchronous parser runner, Libelle needs a clear execution model for job creation, ownership, retries, idempotency, failure persistence, and deployment.

This decision affects:

- public intake response behavior
- `parser_run_id`
- pipeline state transitions
- duplicate prevention
- error visibility
- worker recovery
- staging and Raspberry Pi deployment
- future production hosting

The goal of this issue is to define the implementation contract, not to build the complete worker system.

## Goal

Produce a reviewable architecture recommendation for asynchronous parser execution that preserves the v0.4 trust guarantees:

- no invisible submissions
- no ambiguous pipeline state
- no silent parser failure
- no raw-data overwrite
- traceable processing attempts
- safe retries
- deterministic submission identity

### Intake Transaction Boundary

Define what must complete before the public intake endpoint returns success.

Clarify whether the initial response means:

- request accepted
- submission persisted
- resume persisted
- parser job queued
- parser completed

The intake response must not imply synchronous parser completion if parsing is asynchronous.

### Job Creation

Define:

- what creates the parser job
- when the job is created
- what payload or identifiers it contains
- whether `submission_id` is the canonical job reference
- where queue or job metadata is stored
- how a missing or failed enqueue operation is represented

### `parser_run_id` Ownership

Define:

- which component creates `parser_run_id`
- whether each retry receives a new run ID
- how run IDs relate to one `submission_id`
- how the latest or authoritative parser result is selected
- how prior attempts remain traceable

### Idempotency and Duplicate Prevention

Define:

- the job idempotency key
- how duplicate enqueue attempts are handled
- how duplicate workers are prevented from producing conflicting writes
- what happens when a worker completes after another attempt has already succeeded
- which writes are append-only

### State Transitions

Map the asynchronous flow to the existing state contract.

At minimum address:

- submission persisted
- resume uploaded
- parser pending
- parser running
- parser succeeded
- parser failed
- retry scheduled
- retry exhausted
- resolver pending
- downstream unavailable

Identify which states are persisted versus derived.

### Retry Policy

Define:

- retryable failures
- non-retryable failures
- retry count
- delay or backoff behavior
- manual retry behavior
- how retry exhaustion is exposed in `/snapshot`
- whether a retry creates a new parser result record

### Crash and Restart Recovery

Define behavior when:

- the API process stops after persistence but before enqueue
- the worker stops during parsing
- the Raspberry Pi restarts
- the queue restarts
- a job is claimed but never acknowledged
- storage succeeds but parser-result persistence fails

### Deployment Model

Evaluate a small set of realistic approaches for the current environment, such as:

- database or Sheets-backed job table with polling worker
- Redis-backed queue
- lightweight task framework
- managed queue service
- another minimal durable mechanism

For each serious option, describe:

- operational dependencies
- persistence characteristics
- restart recovery
- local development support
- staging support
- Pi deployment implications
- future migration cost
- observability
- complexity

Do not select technology based only on popularity. Evaluate it against Libelle’s actual scale and deployment model.

### Observability

Define the minimum operational visibility needed:

- queued jobs
- running jobs
- failed jobs
- retry counts
- stale jobs
- processing duration
- correlation by `submission_id` and `parser_run_id`

Avoid logging resume contents or unnecessary personal information.

## Deliverables

Add an architecture document such as:

`docs/architecture/async_parser_execution.md`

The document should include:

1. current synchronous behavior
2. proposed transaction boundary
3. recommended job architecture
4. alternatives considered
5. job and state-transition diagram
6. `submission_id` and `parser_run_id` rules
7. idempotency strategy
8. retry policy
9. crash-recovery behavior
10. local, staging, and production deployment approach
11. security and privacy considerations
12. implementation issue decomposition
13. unresolved decisions or risks

## Acceptance Criteria

- The current parser execution path is documented accurately.
- The intake success boundary is explicitly defined.
- Job creation and persistence rules are defined.
- `parser_run_id` ownership is defined.
- Retry and idempotency rules are defined.
- Duplicate execution behavior is defined.
- Crash and restart recovery are addressed.
- The proposal maps to the existing state-transition contract.
- At least two viable implementation approaches are compared.
- One approach is recommended with clear reasoning.
- Raspberry Pi and staging deployment are addressed.
- Unsupported assumptions and remaining unknowns are stated.
- Follow-up implementation issues are proposed with dependencies and acceptance boundaries.
- No production parser execution behavior is changed in this issue.

## Non-Goals

- Implementing the worker
- Introducing a production queue dependency
- Rewriting the parser
- Changing resolver behavior
- Replacing the state contract
- Building reviewer retry controls
- Adding semantic evaluation
- Performing a production migration

jliu233:
## Summary

Closes #338.

This PR adds a design-only architecture document defining Libelle’s asynchronous parser execution model. It does not change production parser behavior.

The new document covers:

- current synchronous/background-task parser behavior
- proposed public intake success boundary
- durable parser job creation rules
- `submission_id` and `parser_run_id` ownership
- idempotency and duplicate worker behavior
- retry policy and retry exhaustion
- crash/restart recovery
- mapping to the existing v0.4 state contract
- deployment tradeoffs for local, staging, Raspberry Pi, and future production hosting
- observability, security, privacy, and follow-up implementation issues

## Why

The v0.4 plan moves parser execution out of the public intake request path. Before implementing a worker, Libelle needs a clear contract for what “accepted” means, how parser jobs are persisted, how retries are tracked, and how failures remain reviewer-visible.

This document preserves the v0.4 trust guarantees:

- no invisible submissions
- no ambiguous pipeline state
- no silent parser failure
- no raw-data overwrite
- traceable parser attempts
- safe retries
- deterministic submission identity

## Recommendation

The document recommends starting with a Sheets-backed `parser_jobs` tab plus a polling worker.

This approach fits Libelle’s current operating model because it:

- reuses existing Google Sheets/Drive dependencies
- works in local development and staging without a new queue service
- is simple to run on the Raspberry Pi with systemd
- keeps operational state visible in the same place as submissions and parser results
- can later migrate to Redis, Postgres, or a managed queue through a narrow queue abstraction

Alternatives considered include:

- Redis-backed queue
- managed queue service

### Intake boundary

A public intake success response should mean:

- request accepted
- submission validated
- immutable `submissions` row persisted
- resume upload outcome persisted
- durable parser job created, or enqueue failure visibly persisted

It must not imply parser completion.

### Job identity

- `submission_id` remains the canonical submission and logical parser job identity.
- Initial parser job idempotency key is `parse_resume:{submission_id}`.
- Duplicate enqueue attempts should return the existing job instead of creating another active job.

### `parser_run_id`

- The parser worker should own `parser_run_id`.
- Each automatic or manual retry receives a new `parser_run_id`.
- One `submission_id` may have many parser attempts.
- Successful attempts append rows to `parser_results`.
- Failed attempts append traceable error records.

### Retry and recovery

The proposal defines:

- default max attempts of 3
- backoff behavior
- retryable vs non-retryable failures
- stale lease recovery
- API crash recovery after submission persistence
- worker restart behavior
- Raspberry Pi restart behavior
- recovery for uploaded submissions missing parser jobs

## Files Changed

- `docs/architecture/async_parser_execution.md`

## Non-Goals

This PR intentionally does not:

- implement the parser worker
- introduce a production queue dependency
- rewrite the parser
- change resolver behavior
- replace the state contract
- build reviewer retry controls
- perform a production migration
- change public intake runtime behavior

## Testing

Documentation-only change.

No automated tests were run because no runtime code changed.

## Follow-Up Work

The document proposes implementation issues for:

1. adding `parser_jobs` schema and repository
2. moving `parser_run_id` ownership to the worker attempt boundary
3. replacing FastAPI `BackgroundTasks` scheduling with durable enqueue
4. adding the polling parser worker
5. adding a recovery scanner
6. surfacing job observability in `/snapshot` or an internal ops endpoint
7. adding staging and Raspberry Pi worker service definitions

### DIFF
--- a/docs/architecture/async_parser_execution.md
+++ b/docs/architecture/async_parser_execution.md
@@ -44,6 +44,8 @@ Current storage facts:
 - `/snapshot` derives reviewer-facing state from `submissions`,
   `parser_results`, `ops`, and `errors`.
 - The latest parser result is selected by `created_at`, then `parser_run_id`.
+  That is the current v0.4 read behavior, not the target authority rule once
+  durable parser jobs exist.
 
 ## Proposed Intake Boundary
 
@@ -101,6 +103,8 @@ Add a required `parser_jobs` tab:
 | `locked_at` | Claim timestamp. |
 | `lock_expires_at` | Stale-lock recovery timestamp. |
 | `last_parser_run_id` | Latest attempt run ID, successful or failed. |
+| `authoritative_parser_run_id` | Successful attempt that `/snapshot` should read for reviewer-facing parser output. Empty until parser success is finalized. |
+| `parser_started_at` | Timestamp of the first attempt that actually began parser execution. Empty while queued before any parser attempt starts. |
 | `last_error_code` | Most recent failure code. |
 | `last_error_summary` | Short operational failure summary. |
 | `created_at` | Job creation timestamp. |
@@ -125,7 +129,8 @@ Rules:
 
 - Create a parser job only after the `submissions` row has been appended with
   `resume_status == uploaded`.
-- Use `submission_id` as the idempotency key for the initial parser job.
+- Use `parse_resume:{submission_id}` as the idempotency key for the initial
+  parser job.
 - The durable job payload should contain identifiers, not resume contents.
 - Do not store raw resume text in the job table.
 - If text extraction remains in intake temporarily, pass the text to the worker
@@ -136,6 +141,21 @@ Rules:
 - A recovery scanner must be able to create a missing job for any
   `submissions.resume_status == uploaded` record with no parser job and no
   successful parser result.
+
+Submission persistence and parser job creation are not one transaction in
+Sheets. Intake/job creation is therefore eventually reconciled, not atomic.
+The persisted uploaded submission is durable evidence that recovery can use
+even if both job creation and enqueue error logging fail.
+
+The recovery scanner must deterministically repair records matching all of
+these conditions:
+
+- `submissions.resume_status == uploaded`;
+- no `parser_jobs` row exists for `parse_resume:{submission_id}`; and
+- no successful parser result exists for the submission.
+
+The scanner must use the same job idempotency key as intake so repeated scans
+return the existing job instead of creating duplicate active work.
 
 ## parser_run_id Ownership
 
@@ -150,9 +170,14 @@ Rules:
 - Successful attempts append rows to `parser_results`.
 - Failed attempts append rows to `errors` with both `submission_id` and
   `parser_run_id` once the errors schema is extended.
-- The authoritative parser result remains the latest successful
-  `parser_results` row selected by `created_at`, then `parser_run_id`, until a
-  future schema adds an explicit result state.
+- The authoritative parser result is the successful `parser_run_id` recorded in
+  `parser_jobs.authoritative_parser_run_id`.
+- `/snapshot` should read parser-owned and resolver-owned fields from that
+  authoritative attempt. It must not select an older or stale duplicate attempt
+  merely because that attempt appended a later timestamp.
+- `created_at`, then `parser_run_id`, may remain a temporary fallback only for
+  legacy submissions that do not yet have a `parser_jobs` row or
+  `authoritative_parser_run_id`.
 - Prior attempts remain traceable through append-only `parser_results`,
   `errors`, and `parser_jobs` state history or job audit rows.
 
@@ -163,6 +188,18 @@ writer may keep a fallback only as a defensive guard, not as the owner.
 ## Idempotency and Duplicate Prevention
 
 Job idempotency key: `parse_resume:{submission_id}`.
+
+Initial guarantee: duplicate execution may occur; duplicate authoritative state
+must not. Google Sheets does not provide an atomic compare-and-swap claim
+primitive, so two pollers can both read a job as claimable before either sees
+the other's update.
+
+The first Sheets-backed implementation supports one active polling worker in
+staging and Raspberry Pi deployments. Leases exist for crash recovery, stale
+claim detection, and future queue migration; they are not a complete horizontal
+worker concurrency guarantee on Sheets. Running multiple active pollers against
+the same Sheets queue is unsupported until the queue backend provides an atomic
+claim primitive or an equivalent repository-level guarantee.
 
 Duplicate enqueue:
 
@@ -188,11 +225,65 @@ Duplicate worker execution:
   should be monotonic: `succeeded` is terminal unless a reviewer explicitly
   requests a new manual retry.
 
-For Sheets, exact compare-and-swap is limited. The first implementation should
-minimize duplicate claims with a short polling interval, process-level worker
-identity, and a re-read before finalization. If duplicate execution still
-occurs, append-only parser results and latest-row selection keep the reviewer
-state deterministic.
+Parser result write idempotency:
+
+- `(submission_id, parser_run_id)` identifies one logical parser attempt/result.
+- Retrying persistence for the same `parser_run_id` must not create a second
+  logical result.
+- If a Sheets append succeeds but the worker times out before receiving
+  confirmation, the retry path must re-read for `(submission_id,
+  parser_run_id)` before appending. If the row already exists, persistence is
+  complete for that logical result.
+- If physical duplicate rows are ever discovered for the same `(submission_id,
+  parser_run_id)`, repository/read-model code must collapse them to one logical
+  result and surface an operational error; duplicates must not produce
+  conflicting reviewer state.
+
+Authoritative finalization:
+
+- A worker may set `parser_jobs.authoritative_parser_run_id` only after it has
+  persisted parser success for the same `parser_run_id` and re-read the job row
+  to confirm the job is still running for that attempt.
+- If the job already has a different `authoritative_parser_run_id`, the worker
+  must treat its own attempt as stale and leave reviewer-facing authority
+  unchanged.
+- A stale worker must never be able to become authoritative after the job has
+  advanced to another successful attempt.
+
+For Sheets, the repository should still minimize duplicate claims with a short
+polling interval, process-level worker identity, and a re-read before
+finalization. Those mitigations are secondary to the authority rule above.
+
+## Parser and Resolver Boundary
+
+Issue #338 is scoped to moving today's asynchronous parser execution behind a
+durable boundary. The first worker may execute the same unit of work that
+`parse_and_update()` performs today: parse the resume, run Resolver V1 over the
+parser output, and persist the resulting read-model data.
+
+That does not redefine parser/resolver ownership or introduce a general
+pipeline orchestrator. Parser-owned fields remain parser output, resolver-owned
+fields remain Resolver V1 normalization, and `/snapshot` continues to derive
+reviewer-facing state from the v0.4 state contract.
+
+The worker contract must keep parser success and resolver failure separately
+observable:
+
+- Once parser execution succeeds, parser-owned output for that `parser_run_id`
+  must be durably persisted or recoverably idempotent before Resolver V1 runs.
+- Resolver failure after parser success must append/log resolver-stage failure
+  evidence and derive `ParserState = succeeded`,
+  `ResolverState = failed`.
+- A resolver failure after parser success must not set `parser_jobs.status =
+  failed` as a parser failure. For the initial parser job, `succeeded` means the
+  authoritative parser output was persisted; resolver failure is represented by
+  resolver state and error evidence unless a later issue introduces a dedicated
+  resolver job.
+- Resolver failure must not collapse successful parser output into a generic
+  parser failure, delete parser output, or cause `/snapshot` to hide valid
+  parser fields.
+- If resolver enrichment is persisted after parser output, it must target the
+  same logical `(submission_id, parser_run_id)` result and remain idempotent.
 
 ## State Mapping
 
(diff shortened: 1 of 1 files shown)


Kevin-TCUS (changes requested):
Hi Jiaqi,

This is a strong architecture document overall. I think the recommendation to start with a Sheets-backed `parser_jobs` table plus a small polling worker is reasonable for Libelle's current scale and Raspberry Pi/staging deployment, and I like that the queue implementation is intended to remain behind a narrow abstraction.

The document also does a good job defining the intake boundary, `submission_id` versus `parser_run_id`, retry visibility, recovery behavior, and the distinction between operational job state and reviewer-facing state.

Before merging, I think we should tighten a few distributed-systems invariants so the implementation issues do not inherit ambiguous behavior.

### 1. Make duplicate execution an explicit expected failure mode

The document currently describes lease-based claiming and worker ownership, but Google Sheets does not provide an atomic compare-and-swap operation.

Two workers can still both read a job as `queued`, both believe they claimed it, and both execute parsing.

Please make the initial guarantee explicit:

> **Duplicate execution may occur; duplicate authoritative state must not.**

For the first Sheets-backed implementation, I think we should either:

- explicitly support only one active polling worker in staging/Pi deployments, while leases handle crash recovery; or
- define the queue as at-least-once execution and make downstream writes/finalization idempotent enough to tolerate duplicate execution.

I would lean toward documenting the single-active-worker limitation for the first implementation and treating horizontal worker concurrency as unsupported until the queue backend provides an atomic claim primitive.

### 2. Define the authoritative parser attempt explicitly

The document currently says the authoritative parser result is the latest successful `parser_results` row selected by `created_at`, then `parser_run_id`.

I don't think timestamp ordering is strong enough once retries or duplicate workers are possible.

For example, an older/stale attempt could finish after a newer successful attempt and become the newest row.

Please define an explicit authoritative-attempt rule, ideally tied to durable job state rather than inferred solely from timestamps. For example, the job could identify its authoritative successful `parser_run_id`, and `/snapshot` could follow that attempt.

The important invariant is that a stale worker must never be able to become authoritative after the job has advanced to another successful attempt.

### 3. Define parser-result write idempotency

There is another duplicate path if a `parser_results` append succeeds but the worker times out before receiving confirmation.

On retry, the worker could append the same result again.

Please define the persistence idempotency rule for parser results. I think `(submission_id, parser_run_id)` should uniquely identify one parser attempt/result, and retrying persistence for the same run ID should not create another logical result.

This is separate from the job idempotency key `parse_resume:{submission_id}`.

### 4. Preserve parser success when Resolver execution fails

The state model correctly allows:

- `ParserState = succeeded`
- `ResolverState = failed`

but the proposed worker flow currently groups:

`parse -> resolve -> append parser_results`

into one attempt.

Please define how successful parser output remains distinguishable if Resolver V1 fails afterward. A resolver failure should not collapse successful parsing into an undifferentiated parser failure.

The exact persistence ordering can be left to the implementation, but the architecture should guarantee that parser success and resolver failure remain separately observable and consistent with the existing v0.4 state contract.

### 5. Formalize submission/job reconciliation

Appending the `submissions` row and creating the `parser_jobs` row are not one transaction in Sheets.

The recovery scanner already gives us the right model; I would make that explicit:

> **Intake/job creation is eventually reconciled, not atomic.**

Please state the authoritative reconciliation condition clearly, for example:

```text
resume_status == uploaded
AND no parser job exists
AND no successful parser result exists
```

The scanner should deterministically repair that set using the same job idempotency key.

This also clarifies the case where submission persistence succeeds but job creation and error logging both fail: the persisted uploaded submission itself remains sufficient durable evidence for reconciliation.

### 6. Tighten retry/state semantics

A few state descriptions are still ambiguous, particularly:

```text
retry_scheduled -> ParserState = not_started or started
```

For an architecture contract, I think we should define this deterministically based on whether an attempt actually started.

Similarly, please clarify that `max_attempts = 3` means **three total executions**, not an initial execution plus three retries, and make the backoff example consistent with that definition.

### 7. Clarify the parser/resolver worker boundary

The issue is scoped as asynchronous parser execution, while the proposed worker runs both parsing and Resolver V1.

If the intention is simply to move today's existing `parse_and_update()` unit of work behind a durable boundary, please state that explicitly.

I want to avoid this architecture PR unintentionally redefining parser/resolver ownership or introducing a broader pipeline orchestrator beyond the scope of #338.

I don't think these require changing the recommended technology or expanding this into an implementation PR. The Sheets-backed worker remains a reasonable first step.

The main invariant I want the architecture to establish before we implement it is:

> **Duplicate work may happen; duplicate or stale work must never create conflicting authoritative reviewer state.**

Once those authority, idempotency, and reconciliation rules are explicit, I think this will be a very strong foundation for the implementation issues that follow.

Nice work on a difficult architecture problem.

jliu233:
Hi Kevin, thank you for these feedback! I revised the architecture doc to make that explicit and tightened the contract in the areas you flagged:

- documented the first Sheets-backed implementation as supporting one active polling worker, with leases for crash recovery rather than true horizontal concurrency;
- changed authoritative parser result selection to follow `parser_jobs.authoritative_parser_run_id`, with timestamp ordering only as a legacy fallback;
- defined `(submission_id, parser_run_id)` as the logical parser result idempotency key;
- clarified that parser success must remain separately observable when Resolver V1 fails;
- formalized intake/job creation as eventually reconciled, including the recovery scanner condition;
- made retry state mapping deterministic and clarified that `max_attempts = 3` means three total executions;
- scoped the worker boundary to moving today’s `parse_and_update()` unit behind a durable parser job, without redefining parser/resolver ownership.

I also updated the implementation-issue decomposition and risks so the follow-up implementation work inherits those constraints directly.

Kevin-TCUS (changes requested):
Hi Jiaqi,

Thank you for the revision. I went back through the updated architecture document against the concerns from my previous review, and I think you addressed them very well.

In particular, the explicit single-active-worker limitation, authoritative `parser_run_id`, `(submission_id, parser_run_id)` result idempotency, parser/Resolver separation, eventual submission/job reconciliation, deterministic retry state, and narrower worker boundary all make the design substantially stronger.

I have one remaining distributed-systems concern before I think this is ready.

### Lease expiry does not yet guarantee that a stale worker cannot become authoritative

The document now establishes the important invariant:

> A stale worker must never be able to become authoritative after the job has advanced to another successful attempt.

I agree with that invariant, but I don't think the currently described Sheets-backed mechanism fully guarantees it.

The first implementation intentionally uses leases for crash recovery while also acknowledging that Google Sheets does not provide an atomic compare-and-swap claim or finalization primitive. The proposed mitigation is a re-read before finalization.

Consider the following sequence:

```text
Worker A claims job
    |
    | execution exceeds lease timeout
    v
Lease expires

Worker B claims the same job
    |
    | succeeds
    v
B makes parser_run_id B authoritative

Worker A finishes
    |
    | re-reads job
    v
A attempts finalization
```

The re-read helps detect that the job has advanced, but without an atomic conditional write or fencing mechanism there is still a race between the re-read and the final write. In other words, the re-read reduces the race window but does not establish a hard stale-worker guarantee.

I don't think this requires changing the recommendation to use a Sheets-backed queue or introducing Redis/Postgres for this issue. However, I do think the architecture should make the guarantee explicit.

For example, I think one of the following should be documented:

- the first implementation treats lease expiry as an operational boundary, and a worker whose lease has expired must abort without performing further persistence or authoritative finalization;
- authoritative finalization requires a fencing/attempt token that the persistence layer can validate before accepting the write; or
- the document explicitly states that Sheets cannot provide a hard stale-worker fencing guarantee, so the first implementation supports only one active polling worker and multi-worker execution remains unsupported until the queue backend provides an atomic claim/finalization primitive.

The important distinction is:

> Duplicate execution may occur; an expired or stale worker must not be able to overwrite authoritative state established by a newer attempt.

I think the same principle applies to the `(submission_id, parser_run_id)` idempotency rule. It correctly defines the logical identity of a parser result, but Sheets itself cannot atomically enforce physical uniqueness if two racing executions append the same logical result. The architecture should make clear that this is a logical contract rather than a storage guarantee.

Everything else from my previous review looks addressed, and I think this document is now very close. I mainly want the architecture's stated guarantees to match what the underlying storage layer can actually enforce.

Nice work tightening this one up.

jliu233:
Hi Kevin, thank you for your feedback! I revised the doc to make the storage boundary explicit:

- lease expiry is now defined as an operational boundary;
- a worker whose lease has expired must abort before any further durable writes, including parser output, resolver enrichment, error writes for that attempt, or authoritative finalization;
- workers must re-check lease ownership before every durable side effect, not just before finalization;
- the doc now says plainly that Sheets does not provide a hard stale-worker fencing guarantee under multiple active pollers;
- hard multi-worker safety is deferred until the queue backend/repository can atomically validate an attempt or fencing token during claim and finalization;
- `(submission_id, parser_run_id)` idempotency is now described as a logical repository/read-model contract, not a physical uniqueness guarantee that Sheets can enforce atomically.

So the first implementation remains Sheets-backed and single-active-worker, with expired workers required to stop mutating state. Multi-worker execution stays unsupported until we have real atomic fencing.

Kevin-TCUS (approved):
Thanks for the revisions, Jiaqi.

I went back through the updated document against my previous review comments, and I think you've addressed the remaining concerns well.

In particular, I like that the architecture now clearly distinguishes between the logical guarantees we want to uphold and the guarantees that the first Sheets-backed implementation can realistically provide. The document now explicitly:

- limits the initial implementation to a single active polling worker;
- treats lease expiry as an operational boundary;
- requires stale workers to abort before further durable persistence;
- requires re-validation of lease ownership before durable side effects;
- clarifies that `(submission_id, parser_run_id)` is a logical idempotency contract rather than a physical uniqueness guarantee enforced by Sheets; and
- states that hard multi-worker fencing requires a storage backend capable of atomic claim/finalization semantics.

Those changes make the architecture much more consistent with the capabilities of the underlying storage layer while preserving the reviewer-facing trust guarantees that v0.4 is trying to establish.

I also like that the implementation decomposition now carries those invariants forward into the follow-up implementation work rather than leaving them implicit.

Nice work tightening this up. I think this provides a solid foundation for the implementation issues that follow.

**Approved.**

PR #355, opened 2026-08-07
Libelle Product requirements discussion

vinay-personal-troubleshoot:
This PR is an ongoing discussion that validates existing product and tech decisions, and attempts to work back from the customer problem that can serve as guidance to decide future feature prioritization and development.


Kevin-TCUS (commented):
@vinay-personal-troubleshoot Thank you for putting this together. I genuinely enjoyed reading it.

I think you're starting from exactly the right place. Rather than asking "what feature should we build next?", you're asking "what problem are we actually trying to solve for a nonprofit?" I think that's the discussion we should be having as we start looking beyond the current implementation work.

A few things that stood out to me:

- I really like the distinction between **capability** and **intent**. A resume tells us something about what someone can do, but it tells us very little about why they want to contribute, what they actually want to work on, or whether they're a good fit for a particular need. I think that's an important product insight.

- The question **"What is the definition of a successful volunteer found?"** may be the most important one in the document. Before we build search, matching, recommendations, or ranking, we should be clear about what success actually means. Is success someone who applies? Someone we interview? Someone who contributes for six months? The answer changes the product.

- I also like that you're working backwards from nonprofit pain points rather than from existing implementation. That's healthy. It helps us validate why a feature should exist instead of simply extending what we've already built.

A couple thoughts on direction:

I actually don't think of this as a roadmap yet. To me this reads more like the beginning of a **Product Discovery / Requirements** document, and I think that's perfectly okay. I would almost rather let the roadmap emerge from answering these questions than start with a roadmap first.

I'd also tighten the initial scope a little. Today I think Libelle's immediate mission is narrower than "connecting volunteers to nonprofits." Right now we're trying to make TCUS's volunteer intake, understanding, review, and coordination process more trustworthy and scalable. If that generalizes well to other nonprofits later, great—but I'd like us to validate the workflow here first.

One additional question I'd love to explore somewhere in this document is:

> **What is the most expensive human decision Libelle is trying to improve?**

I suspect the answer to that question will influence a lot of what we choose to build (and just as importantly, what we choose not to build) between v0.6 and v1.0.

Finally, I think this would be a great living document. Rather than trying to answer everything immediately, I'd like to gradually replace open questions with explicit product decisions as we learn more. Over time I could imagine sections like:

- Current assumptions
- Open questions
- Product decisions
- Future hypotheses

Overall, I think this is a really valuable direction. Please keep building it. I see this as the start of an ongoing product discussion rather than something that needs to be "finished" in one PR.

PR #358, opened 2026-08-13
[Benchmark] Version annotation derivation independently of synthetic generation

briannadelariva:
## Summary

Closes #347.

Introduces a versioned annotation derivation interface so the synthetic benchmark generator's `gold.json` schema can evolve independently of profile generation, rendering, and templates, per Spike #191's original separation.

- `backend/benchmarks/synthetic/generator/schema.py`:
  - `derive_gold_v1(profile)` — the existing `labeling_rules_v1.md` behavior, unchanged (extracted verbatim from the old `derive_gold`).
  - `derive_gold_v2(profile)` — new: derives the canonical V2 shape from `v2_annotation_spec.md`, mapping `Profile.experience` → structured `EXPERIENCE` section items, `Profile.education` → structured `EDUCATION` items, and `Profile.skills`/`Profile.tools` → a plain-string `SKILLS` section, alongside the required V2 top-level fields (`resume_id`, `source_persona`, `persona`, `name`, `email`, `phone`, `location`, `links`, `skills`, `notes`, `sections`).
  - `derive_gold(profile, version="v1")` — dispatches to the right deriver; defaults to `v1` so the existing call site's behavior is unchanged.
- `generate.py` gains `--annotation-version {v1,v2}` (default `v1`), threaded through to `derive_gold()` and recorded in `manifest.json`.
- `backend/benchmarks/synthetic/README.md` documents the new flag and the section-ordering caveat (derivation order, not literal rendered layout, since `Profile` doesn't carry template section ordering).

The synthetic `Profile` remains the single source of truth. No changes to profile generation, Jinja templates, or PDF rendering.

## Test plan

- [x] New tests in `backend/tests/test_synthetic_annotation_derivation.py`:
  - `derive_gold_v1` output is byte-for-byte identical to the old `derive_gold` behavior.
  - `derive_gold(profile)` with no version arg still matches `derive_gold_v1`.
  - `derive_gold_v2` produces correct top-level fields, section mapping/ordering, and omits empty sections.
  - Generated V2 output is fed through the existing canonical V2 validator from #321 (`validate_v2_golden`) and produces **zero validation issues**, confirming V2 output actually conforms to the canonical contract.
  - Unsupported version raises `ValueError`.
- [x] Full backend suite passes (`pytest tests -q` → 293 passed).

Issue #348, opened 2026-07-29
[Benchmark] Add canonical V2 annotation derivation to the synthetic benchmark generator


Kevin-TCUS:
## Context

Spike #191 established a deterministic synthetic benchmark generator that produces parser-consumable PDF resumes and matching benchmark annotations from a shared synthetic `Profile` model.

The current architecture intentionally separates benchmark generation from benchmark annotation:

```text
Catalogs
    ↓
Profile
   ├──→ PDF renderer (Jinja → HTML → PDF)
   └──→ derive_gold() → benchmark annotation
```

The generator currently derives benchmark annotations according to `labeling_rules_v1.md`.

Since the spike was completed, the benchmark architecture has evolved toward a canonical V2 annotation schema (see #320, #321, and #344). The synthetic benchmark generator should evolve alongside the benchmark while preserving its deterministic generation architecture.

---

## Objective

Extend the annotation derivation stage to emit canonical V2 benchmark annotations from the existing synthetic `Profile` model while preserving the deterministic generation architecture.

The synthetic `Profile` should remain the single source of truth for all generated benchmark artifacts.

---

## In Scope

- Design and implement a canonical V2 annotation derivation path.
- Update `backend/benchmarks/synthetic/generator/schema.py` to support V2 annotation derivation alongside (or replacing) the existing `derive_gold()` implementation, as appropriate.
- Generate canonical V2 benchmark annotations from the existing synthetic `Profile`.
- Preserve deterministic generation.
- Preserve the existing Jinja template library.
- Preserve PDF rendering.
- Ensure generated V2 annotations pass the canonical V2 corpus validator defined by #344.
- Update documentation where appropriate.

---

## Out of Scope

- Parser improvements.
- Resolver improvements.
- Benchmark scoring changes.
- Template redesign.
- New synthetic resume layouts.
- LLM- or agent-assisted generation.

---

## Success Criteria

- Synthetic cases produce canonical V2 benchmark annotations.
- Generated V2 annotations pass the canonical V2 corpus validator defined by #344.
- Existing V1 annotation generation continues to function unless V2 is explicitly selected.
- The annotation derivation layer cleanly supports both V1 and V2 outputs.
- Existing deterministic generation remains reproducible.
- Existing rendered PDFs remain unchanged except where required by the annotation contract.

---

## Motivation

Spike #191 intentionally established a clean separation between:

- synthetic profile generation
- artifact rendering
- benchmark annotation derivation

[...]

Issue #347, opened 2026-07-29
[Benchmark] Version annotation derivation independently of synthetic benchmark generation


Kevin-TCUS:
## Context

Spike #191 intentionally separated synthetic benchmark generation from benchmark annotation derivation.

Today, the synthetic generator derives benchmark annotations through a single implementation:

```python
derive_gold(profile)
```

which encodes the annotation contract defined by `labeling_rules_v1.md`.

As Libelle's benchmark schema evolves, the annotation specification should be able to evolve independently of the synthetic generator, renderer, templates, and underlying synthetic `Profile` model.

The synthetic generator should remain responsible for producing deterministic benchmark scenarios. The annotation derivation layer should be responsible for translating those scenarios into a specific benchmark schema.

---

## Objective

Introduce explicit versioning for synthetic benchmark annotation derivation so that annotation specifications can evolve independently from synthetic benchmark generation.

The synthetic `Profile` should remain the canonical intermediate representation (IR) from which all benchmark annotations are derived.

---

## In Scope

- Design a versioned annotation derivation interface.
- Separate annotation specification from synthetic benchmark generation.
- Support both the existing V1 annotation contract and the canonical V2 contract.
- Refactor annotation derivation as necessary while preserving deterministic behavior.
- Update documentation describing the annotation architecture.

Example approaches may include (but are not limited to):

```python
derive_gold_v1(profile)

derive_gold_v2(profile)
```

or an equivalent abstraction that cleanly separates annotation versions.

---

## Out of Scope

- Parser improvements.
- Resolver improvements.
- Template modifications.
- PDF rendering changes.
- New benchmark fields.
- Synthetic resume generation changes.

---

## Success Criteria

- Annotation derivation supports multiple benchmark schema versions.
- V1 annotation generation continues to function without regression.
- V2 annotation generation integrates cleanly with the canonical benchmark architecture.
- Annotation versioning does not affect deterministic synthetic generation.
- The synthetic `Profile` remains the single source of truth.

---

## Motivation

The synthetic benchmark generator and benchmark annotation compiler solve different problems.

The generator creates deterministic benchmark scenarios.

The annotation compiler translates those scenarios into a specific benchmark schema.

Separating these responsibilities allows benchmark schemas to evolve without requiring changes to synthetic data generation, Jinja templates, or PDF rendering. This preserves the clean architecture established during Spike #191 while making future benchmark evolution significantly easier.

---

**Related:** #191, #320, #321, #344

[...]

PR #359, opened 2026-08-13
[Benchmark] Validate canonical V2 synthetic generation end-to-end

briannadelariva:
## Summary

Addresses #348.

**Note: this PR is stacked on #358 (#347)** — it's based on `version-synthetic-annotation-derivation`, not `main`, because it needs `derive_gold_v2`/`--annotation-version` from that unmerged PR. Once #358 merges to `main`, this PR's base should be retargeted to `main` (GitHub usually does this automatically).

#347 added `derive_gold_v2` and proved it against the canonical V2 validator at the unit-test level (an in-memory `Profile` → `validate_v2_golden()`). This PR proves the same claim end-to-end, and fixes a real gap that only shows up when you do:

1. Ran the actual generator with `--annotation-version v2` (`generate.py --seed 42 --count 8 --annotation-version v2`) — produced real rendered PDFs + `gold.json` files, not synthetic Python objects.
2. Fed that real output through the actual canonical V2 validator **CLI** (`python -m backend.benchmarks.v2_evaluation.validate_v2_goldens`) — **8/8 fixtures valid, 0 errors.**
3. Ran `consistency_check.py` (verifies gold's `location.raw` appears in the extracted PDF text) against the V2 output — and found it hardcoded `gold["submission_id"]`, which doesn't exist on V2 goldens (they use `resume_id`). This would `KeyError` on every V2 case. Fixed: it now resolves the case ID from either field.
4. Re-verified V1 (default) generation + consistency check still work unchanged — 5/5 pass, gold shape identical to before #347.

## Test plan

- [x] `backend/tests/test_synthetic_consistency_check.py` — new unit tests for the V1/V2/fallback ID resolution logic.
- [x] Full backend suite passes (`pytest tests -q` → 296 passed).
- [x] End-to-end, non-unit-test run: `generate.py --annotation-version v2` → real PDFs/gold.json → `validate_v2_goldens` CLI → 8/8 valid, 0 errors.
- [x] `consistency_check.py` → 8/8 V2 cases pass; 5/5 V1 (default) cases pass.


Kevin-TCUS (approved):
Hi Brianna,

Nice work on this. I think this closes the remaining gap after #347 by validating the full end-to-end pipeline rather than only the annotation derivation logic in isolation.

I especially like that this exercises the real generation flow (`generate.py --annotation-version v2`), validates the generated artifacts with the canonical V2 validator, and preserves backward compatibility for the existing V1 workflow. Catching and fixing the `consistency_check.py` assumption around `submission_id` vs. `resume_id` is exactly the kind of issue that only shows up through end-to-end validation.

The scope also stays nicely aligned with #348 without expanding into broader benchmark or parser changes.

Approved — nice work!

PR #357, opened 2026-08-13
[Benchmark] Add corpus preflight validation before benchmark execution

briannadelariva:
## Summary

Closes #344.

Adds a preflight validation step that runs before benchmark scoring begins in `scripts/benchmark.py`. Fails fast with clear, per-fixture diagnostics when the corpus is inconsistent, rather than producing misleading results.

- New standalone module `backend/benchmarks/preflight.py` (kept separate from scoring, per the issue's implementation notes, so future corpus checks don't require touching `scripts/benchmark.py`'s scoring path).
- Validates: matching PDF ↔ golden JSON pairs, filename ↔ internal ID (`submission_id`/`resume_id`) consistency, duplicate fixture IDs, malformed JSON, missing required V1 fields (`skills`, `location`), empty corpus, and mixed schema-version detection (warning only, per the issue).
- `scripts/benchmark.py` now runs preflight before scoring and aborts (exit 1) on any error.
- New flags: `--validate-only` (run preflight and exit, no scoring) and `--allow-missing` (downgrade missing PDF/golden pairs to warnings instead of aborting, preserving the old skip-and-warn behavior).
- Output format matches the issue's desired CLI text (`Benchmark preflight` success block / `Benchmark aborted` failure block with `Fixture:`, `Expected ID:`, `Found ID:`).

## Test plan

- [x] New unit tests in `backend/tests/test_benchmark_preflight.py` covering: matched corpus, empty corpus, missing PDF/golden (default abort vs `--allow-missing` warning), duplicate IDs, mismatched internal ID, malformed JSON, missing required V1 fields, mixed schema versions (warning), and report formatting.
- [x] Full backend suite passes (`pytest tests -q` → 298 passed).
- [x] `python scripts/benchmark.py --validate-only` passes cleanly against the real corpus (`backend/benchmarks/resumes` / `golden_json`).
- [x] Full benchmark run (`python scripts/benchmark.py --parsers libelle`) still produces identical output artifacts for a valid corpus — preflight doesn't change scoring behavior.
- [x] Simulated an ID-mismatch corpus to confirm the abort message matches the issue's exact desired format.

Issue #349, opened 2026-07-29
[Benchmark] Validate generated synthetic benchmark artifacts against the canonical benchmark pipeline


Kevin-TCUS:
## Context

Spike #191 established a deterministic synthetic benchmark generator that derives both rendered resumes and benchmark annotations from a shared synthetic `Profile`.

The resulting generation path is conceptually:

```text
Profile
    ↓
Annotation Derivation
    ↓
PDF Rendering
    ↓
Internal Consistency Check
```

The generator’s internal consistency check protects against drift between the synthetic profile, rendered artifact, and derived annotation.

However, internal consistency is not the same as benchmark-contract validity.

A generated PDF and annotation can remain internally consistent while the annotation still violates the canonical benchmark contract—for example through:

- missing required V2 fields;
- malformed `sections[]` objects;
- unsupported schema shapes;
- invalid schema-version metadata;
- fixture ID or filename mismatches;
- other corpus-level contract violations.

Issue #344 defines canonical corpus preflight validation before benchmark execution. Generated synthetic artifacts should pass that same validation boundary before they are treated as benchmark-ready.

---

## Objective

Integrate canonical benchmark validation into the synthetic benchmark generation workflow so that generated artifacts are accepted as benchmark-ready only when they:

1. remain internally consistent with the synthetic source profile and rendered artifact; and
2. satisfy the canonical benchmark corpus contract defined by #344.

The intended pipeline is:

```text
Profile
    ↓
Annotation Derivation
    ↓
PDF Rendering
    ↓
Internal Consistency Check
    ↓
Canonical Benchmark Validator (#344)
    ↓
Benchmark-Ready Artifact
```

---

## Validation Responsibilities

This issue should preserve the distinction between two validation boundaries.

### Synthetic generation correctness

Does the generated PDF remain internally consistent with the annotation derived from the same synthetic profile?

This is the responsibility of the generator’s internal consistency checks.

### Benchmark corpus correctness

Does the generated fixture satisfy the canonical benchmark schema, identity, pairing, and corpus-integrity requirements?

This is the responsibility of the canonical validator defined by #344.

A fixture should not be considered benchmark-ready unless it passes both boundaries.

---

## In Scope

- Invoke the canonical benchmark validator defined by #344 for newly generated synthetic benchmark artifacts.
- Validate generated annotation schema and required fields.
- Validate PDF and annotation fixture pairing.
- Validate filename, fixture ID, and internal ID alignment where required by the canonical contract.
[...]

Issue #344, opened 2026-07-28
[Benchmark] Add Corpus Preflight Validation Before Benchmark Execution


Kevin-TCUS:
## Summary

Implement a benchmark corpus preflight validation step that verifies the integrity and consistency of Libelle's evaluation corpus **before any scoring begins**.

The benchmark should fail fast with clear diagnostics when the corpus is incomplete, inconsistent, or malformed rather than producing misleading evaluation results.

This issue extracts the **Corpus Preflight Validation** workstream from **#322 – Benchmark Reliability and Extensibility** into a focused implementation task.

---

## Why

As Libelle's evaluation framework expands to include:

- V2 structural goldens
- Layout-aware parsing
- Parser vs. resolver diagnostics
- Section-level evaluation
- Regression tracking

…the benchmark becomes increasingly dependent on the quality of the evaluation corpus itself.

Today, malformed fixtures, mismatched IDs, missing files, or schema inconsistencies can produce confusing benchmark behavior or invalidate evaluation results.

Before we improve parser performance, we should first ensure we're measuring against a trustworthy corpus.

---

## Goal

Add a validation stage that executes **before benchmark scoring** and verifies the benchmark corpus is internally consistent.

Validation should include:

- Matching PDF ↔ golden JSON fixture pairs
- Filename ↔ internal fixture ID consistency
- Duplicate fixture IDs
- Malformed JSON
- Missing required fields
- Unsupported schema versions
- Empty benchmark corpus
- Detection of mixed schema versions (emit a warning; enforcement policy can be finalized after the V2 annotation specification stabilizes)

Only after successful validation should benchmark scoring begin.

---

## Desired Behavior

Successful validation:

```text
Benchmark preflight

✓ 20 PDFs found
✓ 20 golden JSON fixtures found
✓ IDs match filenames
✓ No duplicate fixture IDs
✓ Schemas validated

Starting benchmark...
```

Failure example:

```text
Benchmark aborted

Fixture: multi_col_04

Expected ID: multi_col_04
Found ID: multi_col_05
```

---

## Implementation Notes

The current benchmark runner lives in `scripts/benchmark.py`, with benchmark fixtures and supporting artifacts under `benchmarks/`.

This issue should focus on adding a preflight validation layer before scoring begins. Prefer keeping the validation logic separate from the scoring path so additional corpus checks can be added over time without unnecessarily expanding the benchmark runner.

[...]

PR #360, opened 2026-08-13
[Benchmark] Gate synthetic generation on canonical corpus validity

briannadelariva:
## Summary

Addresses #349.

**Note: this PR is 4-deep stacked** — based on `validate-synthetic-v2-generation-e2e` (#359/#348), which is based on `version-synthetic-annotation-derivation` (#358/#347). This branch also locally merges in `add-benchmark-corpus-preflight-validation` (#357/#344) to get `backend/benchmarks/preflight.py`, since #349 explicitly reuses that validator rather than reimplementing it. Suggested merge order: #357 → #358 → #359 → this one, retargeting each PR's base to `main` as the one below it merges (GitHub usually does this automatically).

#349 closes the loop the last three issues built toward: a generated PDF + annotation can be internally consistent (same source profile) while still violating the canonical benchmark corpus contract (missing fields, ID mismatch, malformed JSON, etc). This PR adds a gate so generated artifacts are only reported "benchmark-ready" when they pass **both** boundaries.

- New `backend/benchmarks/synthetic/generator/validate_generated.py` — runs the generator's existing `consistency_check.py` **and** the canonical preflight validator from #344 (`backend/benchmarks/preflight.py`) against a generated corpus, and prints a combined report ending in `BENCHMARK-READY` / `NOT BENCHMARK-READY` (exit 0/1). Reuses both existing validators; doesn't reimplement either.
- `generate.py` gains an opt-in `--validate` flag that runs this automatically after generation and exits non-zero if the corpus isn't benchmark-ready — usable in CI without changing the existing default behavior (flag omitted = unchanged).
- **Real bug fix surfaced while testing**: `run_consistency()` was crashing with an unhandled `JSONDecodeError` on malformed `gold.json` instead of reporting it as a failure. Fixed so malformed input is reported, not fatal.

## Test plan

- [x] `backend/tests/test_synthetic_validate_generated.py` — covers all cases the issue requires: one valid fixture, one malformed annotation, one PDF/annotation pairing failure, one fixture identity mismatch, one unsupported schema-version case — plus an internal-consistency failure case for full boundary coverage.
- [x] Full backend suite passes (`pytest tests -q` → 314 passed).
- [x] Real end-to-end run: `generate.py --annotation-version v2 --validate` against 5 freshly rendered cases → `BENCHMARK-READY`.
- [x] Forced an ID mismatch in generated output and confirmed the gate correctly reports `NOT BENCHMARK-READY` with the exact fixture/expected/found diagnostic from #344's format.

### DIFF
--- a/backend/benchmarks/preflight.py
+++ b/backend/benchmarks/preflight.py
@@ -1,0 +1,226 @@
+"""Corpus preflight validation for the V1 benchmark runner (scripts/benchmark.py).
+
+Validates the resume PDF / golden JSON corpus for internal consistency
+before any benchmark scoring begins (#344). Kept separate from scoring so
+additional corpus checks can be added here without touching the scoring
+path.
+"""
+
+from __future__ import annotations
+
+import json
+from collections import Counter
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Tuple
+
+
+@dataclass
+class PreflightIssue:
+    fixture_id: str
+    message: str
+    severity: str = "error"  # "error" or "warning"
+    expected: Optional[str] = None
+    found: Optional[str] = None
+
+    def format(self) -> str:
+        lines = [f"Fixture: {self.fixture_id}", ""]
+        if self.expected is not None or self.found is not None:
+            if self.expected is not None:
+                lines.append(f"Expected ID: {self.expected}")
+            if self.found is not None:
+                lines.append(f"Found ID: {self.found}")
+        else:
+            lines.append(self.message)
+        return "\n".join(lines)
+
+
+@dataclass
+class PreflightResult:
+    pdf_count: int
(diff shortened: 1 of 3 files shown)


Kevin-TCUS (approved):
Hi @briannadelariva,

Nice work on this. I think this cleanly closes the validation loop that the previous benchmark issues were building toward.

I especially like that this reuses the existing validation boundaries instead of introducing a third implementation. `validate_generated.py` acts as an orchestration layer, combining the generator's internal consistency check with the canonical corpus preflight validator while keeping each responsibility separate.

The additional handling for malformed `gold.json` files is also a nice improvement—it makes the validation boundary more robust without expanding the scope beyond what #349 is trying to accomplish.

Overall this stays well aligned with the issue and preserves the separation between generation, validation, and benchmark execution.

Approved — nice work!

